PR #4, opened 2016-04-11
Fix documentation

postfalk:
In the complex request section, it should be `client.get_type()` instead of  `client.get_element()` method.

I assume it also needs to changed in the section below.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -32,11 +32,11 @@ Complex requests
 ================
 
 Most of the times you need to pass nested data to the soap client. These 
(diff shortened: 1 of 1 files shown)


mvantellingen:
Hi, thanks for the PR! You are right, it's actually now that I think about it a part of the api i'm still not really happy about. Complex types can also be anonymous and then you have to use `get_element('xx').type()` which is not too user friendly.

mvantellingen:
Follow up in #5 let me know if you have an idea ;-)

PR #9, opened 2016-05-01
WSSE support

mvantellingen:
Adds a ws-security implementation:

- [x] x509 certificates using binary security token
- [ ] username/password


mvantellingen:
dm.xmlsec.binding doesn't support python3, other approach is to use xmlsec (see py-wsse)

PR #12, opened 2016-05-07
Allow disabling SSL certificates verification

onjin:
Hi. It would be nice to have control over all `requests` params, but so far I cannot use `zeep` because of SSL verification.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -77,13 +77,15 @@ Caching
 =======
 The default cache backed is SqliteCache.  It caches the WSDL and XSD files for 
 1 hour by default. You can disable caching by passing `None` as value to the
-`cache` attribute when initializing the client::
+`Transport.cache` attribute when initializing the client::
 
     >>> from zeep import Client
     >>> from zeep.cache import SqliteCache
+    >>> from zeep.transports import Transport
+    >>> transport = Transport(cache=None)
     >>> client = zeep.Client(
     ...     'http://www.webservicex.net/ConvertSpeed.asmx?WSDL', 
-    ...     cache=None)
+    ...     transport=transport)
 
 
 Changing the SqliteCache settings can be done via::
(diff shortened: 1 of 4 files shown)


mvantellingen:
hi @onjin, thanks for the PR! 

The changes to the transport module look good, but I'm not sure about the `verify` kwarg in `client.Client()`. 

I see one of the following options:

```
    client = zeep.Client(wsdl, transport_options={'verify': False, 'timeout': 600})
```

Or the more explicit:
```
    transport = zeep.transport.Transport(verify=False, timeout=600)
    client = zeep.Client(wsdl, transport=transport)
```

But you opinion is more then welcome 👍

onjin:
hi @mvantellingen 

explicit versions looks good, but then should we move `cache` option from `Client` to `Transport`?

    transport = zeep.transports.Transport(verify=False, timeout=600, cache=None)
    client = zeep.Client(wsdl, transport=transport)

    class Transport(object):
        def __init__(self, cache=None, timeout=300, verify=True):
            self.cache = cache or SqliteCache()

mvantellingen:
Yeah, agree. The cache is currently only used within the transport layer. My initial thinking was to use it for the parsed wsdl state, but in practice the parsing is fast enough to skip caching here.

onjin:
I removed `cache` option from `Client` because it's still `dev` release. But i can make it backward compatible with deprecation warning if we need that.

mvantellingen:
Thanks! Can you also update the __main__ module? I've just added a unittest for it on the master branch. 

Regarding the deprecation warning, I think we can skip it indeed. After the 1.0 release (which shouldn't take too long) we should aim for good semver practices.

onjin:
Sure. `__main__` updated and pushed.

mvantellingen:
Thanks ! :D

PR #20, opened 2016-05-12
fix SOAP 1.2 content type

benzolius:
Is there a reason for "content_type = 'application/xml+soap; charset=utf-8'"

Here https://www.w3.org/TR/soap12-part1/#reltoxml  I find only:

"application/soap+xml"


mvantellingen:
Thanks!

PR #23, opened 2016-05-13
Fixed examples.

VertigoRay:
Previous method would give you `NameError: name 'zeep' is not defined`, as [I pointed out here](http://stackoverflow.com/a/37207944/615422).


mvantellingen:
Oh thanks, stupid mistake on my part :(

VertigoRay:
It happens to the best of us. :beers:

PR #34, opened 2016-05-24
Use Header even if object does not have a header

cheuel-novero:
* A Header needs to be created if there is not a header already in the
  Element. This was not covered by the old implementation


mvantellingen:
Thanks! Do you think you can add a unittest for this? If not then let me know, i'll pick it up myself

cheuel-novero:
I think you alread picked this up, right?

mvantellingen:
Yeah merged your PR indeed. And altered some logic again this morning :)

PR #62, opened 2016-06-09
Always enclude the EncodingType attribute in the Nonce element

antocuni:
According to the specs [0] the attribute is optional and thus strictly speaking zeep is
correct.  However, according to this stackoverflow answer [1] and to my
experience, some Java Web Services don't work if you don't include the
EncodingType attribute.

[0] https://www.oasis-open.org/committees/download.php/16782/wss-v1.1-spec-os-UsernameTokenProfile.pdf
[1] http://stackoverflow.com/questions/7418506/how-do-i-add-an-encodingtype-attribute-to-the-nonce-element-of-a-usernametoken-i

### DIFF
--- a/src/zeep/wsse/username.py
+++ b/src/zeep/wsse/username.py
@@ -42,7 +42,8 @@ class UsernameToken(object):
         </wsse:Security>
 
     """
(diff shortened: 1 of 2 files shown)


mvantellingen:
nice one, thanks!

PR #67, opened 2016-06-10
Unknown fault should raise `zeep.exceptions.Fault`

erwaller:
Rather than `AssertionError`.

Looks like that assertion may have been left in there accidentally.


mvantellingen:
Oh yeah thanks! I'll merge this after adding a unittest for it

erwaller:
Great, thanks!

PR #73, opened 2016-06-14
implemented suds-like last_sent and last_received methods

adilkhash:
Hi, I decided to impelement simple methods (like in suds) to analyze what has been sent and received to/from web service (XML content). It would be great if you agree on this.

### DIFF
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -100,3 +100,9 @@ class Client(object):
 
     def get_element(self, name):
         return self.wsdl.schema.get_element(name)
+
+    def last_sent(self):
+        return self.transport.last_sent()
+
+    def last_received(self):
+        return self.transport.last_received()
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -16,6 +16,8 @@ class Transport(object):
         self.session = self.create_session()
         self.session.verify = verify
         self.session.auth = http_auth
+        self._sent = None
+        self._received = None
 
     def create_session(self):
         return requests.Session()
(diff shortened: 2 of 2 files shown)


mvantellingen:
Hi @adilkhash, first of all thanks for the PR.

I'm not a fan of the current solution. In my personal use-case I often receive soap responses which are around 40 to 60 mb. Keeping that around by default for potential debugging scenario's isn't feasible.

However we could add a DebuggingTransport class which adds this functionality (and doesn't expose it to the client). B'm currently more interested in the use-case you are trying to solve. Perhaps there is a better more generic way.

adilkhash:
I do not have particular use-case, it is just for debugging purposes, especially with poorly documented web services. I totally agree with you that in case of enormous services my solution is not suitable. Would you like me to impement DebuggingTransport ?

mvantellingen:
Yeah that would be cool! 👍  We could also add a logger.debug() call with the contents which will be sent in that one.

adilkhash:
Btw, when I worked on a daily basis with SOAP services using suds, I had to log request/response payloads in order to do investigations in case of failures (usually I had to deal with payment gateways, external booking services). What do you think the best way to deal with it?

Previously I have used logging module in my client API which logged last_received and last_sent outputs.

mvantellingen:
Well if you create a DebuggingTransport() and do a logging.debug statement (or even a print() statement) in the post() and get() then you have enough info

adilkhash:
No, I mean in production. I had a practice to log everywith while working with external webservice.

mvantellingen:
Perhaps an option to register a callback on the transport?

adilkhash:
OK, I will investigate that option.

mvantellingen:
See http://docs.python-zeep.org/en/master/plugins.html#historyplugin :)

PR #82, opened 2016-06-17
The SOAPAction header value should be double quoted

donkopotamus:
According to the [specification](http://www.w3.org/TR/soap11/#_Toc478383528), the SOAPAction header should be wrapped in double quotes.  Although many services are happy to accept an unquoted SOAPAction header, others with stricter parsing will reject such messages.


mvantellingen:
Cool! Thanks!!

Issue #97, opened 2016-06-23
Using zeep in Django with USE_TZ setting causes "can't compare offset-naive and offset-aware datetimes" error


tkocmathla:
If I use zeep within a Django application that uses sqlite as the backend, and has the USE_TZ setting enabled, zeep throws a timezone offset-related error. 

The datetime coming back from sqlite is timezone-aware, but the datetime stored in `offset` is timezone-naive because [datetime.utcnow() is timezone-naive](https://docs.python.org/2/library/datetime.html#datetime.datetime.utcnow). 

Here's the stack trace:
```
Traceback (most recent call last):
<elided>
  File "/usr/lib/python2.7/site-packages/zeep/client.py", line 45, in __init__
    self.wsdl = Document(wsdl, self.transport)
  File "/usr/lib/python2.7/site-packages/zeep/wsdl/wsdl.py", line 60, in __init__
    document = self._load_content(location)
  File "/usr/lib/python2.7/site-packages/zeep/wsdl/wsdl.py", line 117, in _load_content
    location, self.transport, self._parser_context, self.location)
  File "/usr/lib/python2.7/site-packages/zeep/parser.py", line 44, in load_external
    response = transport.load(url)
  File "/usr/lib/python2.7/site-packages/zeep/transports.py", line 24, in load
    response = self.cache.get(url)
  File "/usr/lib/python2.7/site-packages/zeep/cache.py", line 71, in get
    if not self._timeout or created > offset:
TypeError: can't compare offset-naive and offset-aware datetimes
```

mvantellingen:
Fixed in #98 (thanks again!)

PR #98, opened 2016-06-23
Ensure cache datetimes are timezone-aware (#97).

tkocmathla:
This works for me both outside of and within Django. The UTC class I used is taken from the Python docs here: https://docs.python.org/2/library/datetime.html#datetime.tzinfo.fromutc.

### DIFF
--- a/src/zeep/cache.py
+++ b/src/zeep/cache.py
@@ -2,27 +2,11 @@ import base64
 import datetime
 import errno
 import os
+import pytz
 import sqlite3
 
 import appdirs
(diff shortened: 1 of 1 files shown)


mvantellingen:
Hi, thanks for the PR!. There is a dependency on pytz which includes a pytz.utc timezone. So you can do ``datetime.datetime.utcnow().replace(tzinfo=pytz.utc)``

mvantellingen:
nice! Thanks! 👍

Issue #75, opened 2016-06-14
'Any' object has no attribute 'qname'


c-engelhardt-ecoservice:
I'm trying to send the following request:

```
...
client = zeep.Client('https://adi-uat.apmg-international.com/95/lookupdata.asmx?WSDL')
response = client.service.GetExamName(ATO_ID=ato_id, ATO_KEY=ato_key)
...

```

But get the following issue:

```
Traceback (most recent call last):
  File "soap_connection.py", line 161, in sync_soap_object
    response = client.service.GetExamName(ATO_ID=ato_id, ATO_KEY=ato_key)
  File "/usr/lib/python2.7/site-packages/zeep/client.py", line 24, in __call__
    self._proxy._client, self._op_name, args, kwargs)
  File "/usr/lib/python2.7/site-packages/zeep/wsdl/definitions.py", line 298, in send
    client, self.binding_options, operation, args, kwargs)
  File "/usr/lib/python2.7/site-packages/zeep/wsdl/soap.py", line 99, in send
    return self.process_reply(client, operation_obj, response)
  File "/usr/lib/python2.7/site-packages/zeep/wsdl/soap.py", line 134, in process_reply
    doc = operation.process_reply(doc)
  File "/usr/lib/python2.7/site-packages/zeep/wsdl/soap.py", line 265, in process_reply
    return self.output.deserialize(node)
  File "/usr/lib/python2.7/site-packages/zeep/wsdl/messages.py", line 205, in deserialize
    result.append(part.element.parse(elm, self.wsdl.schema))
  File "/usr/lib/python2.7/site-packages/zeep/xsd/elements.py", line 128, in parse
    return self.type.parse_xmlelement(value, schema)
  File "/usr/lib/python2.7/site-packages/zeep/xsd/types.py", line 255, in parse_xmlelement
    result = field.parse(element, schema)
  File "/usr/lib/python2.7/site-packages/zeep/xsd/elements.py", line 128, in parse
    return self.type.parse_xmlelement(value, schema)
  File "/usr/lib/python2.7/site-packages/zeep/xsd/types.py", line 242, in parse_xmlelement
    while field and field.qname != element.tag:
AttributeError: 'Any' object has no attribute 'qname'
```

I attached the request/response xml from SoapUI as example:
[example.zip](https://github.com/mvantellingen/python-zeep/files/313364/example.zip)

mvantellingen:
[...]

PR #118, opened 2016-07-08
Fix for NoneType error

mattpepin:
Here's the pull request to fix the end of issue #75 :

https://github.com/mvantellingen/python-zeep/issues/75#event-713973411

Thanks :)


mvantellingen:
Thanks for the PR! Really appreciated. However I've decided to try an alternative fix first, by simply not trying to parse types for which there is no attribute. See bf593b7b278a43ceac6fba70b5ae18c3bb526dc2

mattpepin:
I agree. I figured it was getting complicated (adding None checks everywhere). It got me running in the mean time.

PR #122, opened 2016-07-11
Parse XML response even if it has errors

sebasmonia:
Using the parser with recover=True allows XML with errors to be parsed.

If the change is not good, maybe this could be exposed as an option in zeep itself? When instantiating the client, maybe.

### DIFF
--- a/src/zeep/wsdl/soap.py
+++ b/src/zeep/wsdl/soap.py
@@ -110,7 +110,8 @@ class SoapBinding(Binding):
                 % response.status_code)
 
         try:
-            doc = fromstring(response.content)
+            parser = etree.XMLParser(recover=True)
+            doc = fromstring(response.content, parser)
         except etree.XMLSyntaxError:
             raise TransportError(
                 u'Server returned HTTP status %d (%s)'


sebasmonia:
Let me know if you want the "option" approach. I can try to work on that.

mvantellingen:
Hi, thanks for the PR. I'm indeed not sure if this is something we want to enable by default. Reason for this is that I don't want to silently accept invalid data. An option is an idea, but perhaps in that case we need to create a generic option interface.

I'm wondering what the exact use-case is. Are there encoding issues or is the xml really invalid?

sebasmonia:
Creating a generic option interface sounds like a good idea. It could also be an attribute in the client, if there aren't enough options to justify that change:
c = Client(url)
# if you really need invalid stuff to go through...
c.attempt_to_parse_invalid_xml = True # not the best name
I was getting the exception in the Transport, so maybe the option should live there.

Regarding my XML, I'm pretty sure it was just encoding but there was also a chance that .NET serializers are adding something invalid for newlines or something like that...I took this as a shortcut so that I wouldn't spend more time dealing with something that might or might not be a problem.
That's why I suggested the option approach.

mvantellingen:
Ok thanks. So I'm not in favour of adding to many options since that adds to the maintenance burden.  Are you able to supply the response which resulted in the error so that I can see if this is solvable in a generic way?

sebasmonia:
Someone changed something in the item and it doesn't break anymore. If I run into this again I'll share the details with you (or even better, get a real solution!).

mvantellingen:
Ok thanks, closing it for now

PR #131, opened 2016-07-18
Add formatting options to XML posted via the transport class

fiebiga:
If you'd rather not add the change, I understand - we just have to consume a (bad) "SOAP" api that requires non-pretty_print XML or it spits back a 500, so this option is useful for us.

### DIFF
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -19,9 +19,10 @@ class OperationProxy(object):
         self._op_name = operation_name
 
     def __call__(self, *args, **kwargs):
+        pretty_print = kwargs.pop('pretty_print', True)
         return self._proxy._binding.send(
             self._proxy._client, self._proxy._binding_options,
-            self._op_name, args, kwargs)
+            self._op_name, args, kwargs, pretty_print=pretty_print)
 
 
 class ServiceProxy(object):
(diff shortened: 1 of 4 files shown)


mvantellingen:
Thanks. I think a better method which I can support is handling this in the transport class. So we need to add a post_xml() method to the transport class and we serialize the xml node there instead of in the binding. This gives the ability to set the serialization settings in the transport class.

fiebiga:
Thanks for the quick response!

Yeah I didn't think this was the most appropriate place to put this kwarg,
but I was trying to be as unobtrusive as possible with the change. Shall I
alter it in that manner or were you going to?

mvantellingen:
If you want to pick it up that would be great 👍  Otherwise I can take a look later this week.

fiebiga:
Sounds good - I'll take a look, thanks!

fiebiga:
Ok, how about that? I took it a step farther and allowed someone to pass all formatting options. I went back and forth on whether to use **post_xml_formatting_options in the __init__, or to relegate it to a dict, but landed on the former

mvantellingen:
Hi. I've implemented this in master with additional changes. I've decided this scenario is an edge case and that adding additional options to alter the serialisation leads to additional maintenance burden. 

So the proper way to change the serialisation is to subclass the transport class and override the post_xml() method and do the serialisation yourself there. 

Thanks again and hope you can agree with my reasoning

fiebiga:
Yes, this is certainly an edge case, just as finding a whitespace sensitive "soap" api is an edge case lol. Having post_xml to override in a subclass is all we need, thanks!

PR #144, opened 2016-07-23
Gracefully handle Invalid times (xsd) values by returning None instead of raising an exception

js1972:
I've noticed on some web services (particularly those exposed by SAP systems) that have xsd time values, return "  :  :  " as an empty value - which python (isodate module) treats as invalid and raises an exception from the builtins.py Time class. This exception is unhandled by zeep so the entire parsing operation fails because of these times.

So in this pull request I've simply handled the exception and return None.

This may not be the best place to do this architecturally as I'm new to Python - but it works...  ;-) and I can now work with SAP web services.


mvantellingen:
Fixed it via 48b2b21af9b674c7aa957a09c94a2ddb169d55bf 

Thanks again for the PR though! :-)

PR #148, opened 2016-07-26
Add Client X509 Certificate Option to Transport

techguy613:
- Add local certificate keyword argument to Transport.**init**() - MSSL authentication
- Update Transport.create_session() to accept a cert argument and set the Session()'s cert value

Allow the zeep client to provide a local certificate as per the `cert` argument in the `requests` module.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 /frutsels/
 /server/
 /htmlcov/
+.idea/
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -10,20 +10,23 @@ from zeep.wsdl.utils import etree_to_string
 
 class Transport(object):
 
-    def __init__(self, cache=NotSet, timeout=300, verify=True, http_auth=None):
+    def __init__(self, cache=NotSet, timeout=300, verify=True, http_auth=None, cert=None):
         self.cache = SqliteCache() if cache is NotSet else cache
         self.timeout = timeout
         self.verify = verify
         self.http_auth = http_auth
         self.logger = logging.getLogger(__name__)
-        self.session = self.create_session()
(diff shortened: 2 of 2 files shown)


mvantellingen:
Thanks! I really need to think this through some more. There are a lot of requests for passing additional requests settings through the transport class. I don't want to get a situation where the transport's ``__init__`` taks 10+ kwargs. perhaps two attributes: ``extra_requests_kwargs`` and ``extra_requests_attributes`` would make sense?

techguy613:
Agreed, let me make that change.

techguy613:
Hey @mvantellingen, I made the update, but I decided to just include **kwargs to take any additional arguments so it's easier to just pass them in on the __init__() call. In this way, what you do with them and what they affect will be up to whoever is making the update.

mvantellingen:
Thanks for your work! I will however try to come up with a more generic method myself in the next few days.

techguy613:
Thanks @mvantellingen! I would really like to use your module for a project I'm working on, but sadly I need client cert capabilities. Hopefully see some changes soon :) 👍

mvantellingen:
You can already use it of course, by simply subclassing the transport class, adding the cert to it and passing that to the client. This is one of the supported methods.

techguy613:
Of course, it would just love to see the option in the top-level transport itself :) Thanks @mvantellingen!

Untou4able:
How do you feel about adding session karg to Transport class?

mvantellingen:
@Untou4able: Great idea actually. Just did that, see http://docs.python-zeep.org/en/master/transport.html#http-authentication

techguy613:
Thanks @mvantellingen! 👍

PR #156, opened 2016-08-01
Fix any simple type parsing

m-vellinga:
Parsing of a array of simple types failed. Not sure if this is the right way to have it fixed since the items in the array are parsed as a Any element apparently.

If you have any other suggestions for a better solution let me know and i'll make it happen.


mvantellingen:
Looks good on first sight! Thanks! 

To fix the PY3 errors see https://github.com/mvantellingen/python-zeep/commit/8a7ccffb547094401d31dada4c00681a3f54682f

m-vellinga:
I'm going to create a new pull requests, I see I kinda screwed up the history and have the py3 things fixed now I guess. Thanks for the pointer.

mvantellingen:
Thanks. In the future you can just do 'git rebase -i origin/master' and squash all commits and then force push. But this will work as well :-)

Issue #164, opened 2016-08-12
Namespace prefixes ignored on SOAP responses.


anaoum:
In certain implementations, the server will ignore the namespace prefixes used in the SOAP request, and use its own prefixes in the `xsi:type` tags of the response. This results in the response failing to parse because `zeep.xsd.Schema` stores its own prefix map and uses this when creating a QName (see `Schema._create_qname()`).

For example, in the following response XML:

```
<?xml version="1.0" encoding="utf-8"?>
<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:SOAP-ENC="http://schemas.xmlsoap.org/soap/encoding/" xmlns:tns="http://api.radioreference.com/soap2">
    <SOAP-ENV:Body>
        <ns1:getTagResponse xmlns:ns1="http://api.radioreference.com/soap2">
            <return xsi:type="SOAP-ENC:Array" SOAP-ENC:arrayType="tns:tag[1]">
                <item xsi:type="tns:tag">
                    <tagId xsi:type="xsd:int">4</tagId>
                    <tagDescr xsi:type="xsd:string">EMS Dispatch</tagDescr>
                </item>
            </return>
        </ns1:getTagResponse>
    </SOAP-ENV:Body>
</SOAP-ENV:Envelope>
```

The exception is thrown: `ValueError: No namespace defined for the prefix 'tns'` because this namespace was defined as `ns0` by the Schema in `Schema._create_prefix_map()`.

mvantellingen:
Thanks, #184 turned out to be a duplicate of this afaik.

PR #165, opened 2016-08-12
Fix issue #164 by using response namespace map

anaoum:
This code will use the xml namespace map provided in the SOAP response
from the server to fully expand the namespace prefixes used in the
xsi:type tags. This ensures that namespace prefixes used by the server
are expanded using the server generated xmlns tags rather than relying
on the request prefixes to remain the same.

### DIFF
--- a/src/zeep/xsd/elements.py
+++ b/src/zeep/xsd/elements.py
@@ -78,6 +78,10 @@ class Any(Base):
 
         xsd_type = xmlelement.get('{http://www.w3.org/2001/XMLSchema-instance}type')
         if xsd_type is not None:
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks for the PR! Can you add a unittest for this?

mvantellingen:
Hi @anaoum, I think i solved the issue via https://github.com/mvantellingen/python-zeep/commit/d44f3fc73f6f2a9b567be3ece077d018cc49cfa6#diff-7f3a571ca715a309fee30bd8e450cce6R85 

Thanks for the PR, really appreciated!

Issue #140, opened 2016-07-20
Transport timeout option not functioning properly


ryan-mccaffrey:
I've tried setting the timeout option in the Transport constructor to timeout=10 and simulated a timeout by disabling my wifi. I found that it took about 3 minutes for the request to return a ReadTimeout error. Trying a second time, the error came back after 1 minute, which is not only the 10 seconds I specified, but also inconsistent with the 3 minutes from before.

My code is as below:
```
transport = Transport(http_auth=(username, apiKey), timeout=10)
connection = Client(wsdl=myURL, transport=transport)
```

If I'm doing something wrong, any tips would be greatly appreciated. Thanks!

mvantellingen:
Seems that the timeout for now is only used for loading wsdl/xsd files. Not for XML POST/GET requests. That should probably be added (and perhaps as another kwarg).

ryan-mccaffrey:
Speaking of POST/GET requests, is there a way to access the properties of the HTTP responses from the requests? I'm currently trying to figure out how to get the response's status code.

mvantellingen:
You can achieve this by subclassing the tranport class and keeping a reference to the last response object there. It's discussed a bit in #91.

beager:
+1 to this, XML POST/GET should use the timeout given. Less importantly, being allowed to specify timeouts on a per-binding or per-request basis might be good.

mvantellingen:
This can now be done via::
```python
transport = Transport(operation_timeout=360)
client = Client(url, transport=transport)
```

Or 
```python
client = Client(url)
with client.options(timeout=360):
    clients.service.slow_call()
```

PR #172, opened 2016-08-29
Add support for Transport-specified timeout in XML GET/POST requests

beager:
This fixes issue #140 which disregards the timeout passed to `Transport` for any request except the one that fetches the wsdl/xsd.

This has some weird considerations in that by default the wsdl/xsd fetch is 300s, but there is no timeout set for get/post requests. Any "reasonable" request would be much less than that, but you can't go about assuming that people won't be making a 2000s request. As such, I set the timeout by default to `NotSet` and assigned it to `300` for `load()` and `None` for get/post if it was not set on instantiation of the Transport.

This could potentially break implementations where a load timeout is explicitly set, but the get/post request should not have a timeout. Not sure what your thoughts on that are.

Anyway, this would be hugely useful for me to get this merged. Not familiar with your testing/coverage concerns but tox seems to pass for this. Thanks.

@mvantellingen

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -52,15 +52,17 @@ class Transport(object):
         with open(url, 'rb') as fh:
             return fh.read()
 
-    def post(self, address, message, headers):
+    def post(self, address, message, headers, timeout=NotSet):
+        timeout = self.timeout if timeout is NotSet else timeout
         self.logger.debug("HTTP Post to %s:\n%s", address, message)
-        response = self.session.post(address, data=message, headers=headers)
+        response = self.session.post(
+            address, data=message, headers=headers, timeout=timeout)
         self.logger.debug(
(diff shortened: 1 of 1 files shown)


adamcavendish:
That's great! Hope that it'll get to work asap.

mvantellingen:
I still have to look better at this, but maybe we need to split the timeout settings? Like call_timeout=None, wsdl_timeout=60 or something like that? We might then also want to support changing the timeout per call with something like:

with client.transport.timeout(60):
      client.service.call_something_which_needs_to_return_in_60()

mvantellingen:
See 908ed9e32eb66720b9df0c60efe4a8be33b42e66 and 9c5510c4d638774d9ce2454e2055d81de3f5d4bb

beager:
hooray, thank you @mvantellingen 🎉

PR #185, opened 2016-09-15
Add try catch block when accessing node.getchildren()[0]

andrewserong:
I ran into an `IndexError: list index out of range` with a fairly complex WSDL that I don't have any control over. Adding this check got it working for me.

### DIFF
--- a/src/zeep/wsdl/http.py
+++ b/src/zeep/wsdl/http.py
@@ -143,16 +143,20 @@ class HttpOperation(Operation):
                 continue
 
             # XXX Multiple mime types may be declared as alternatives
-            message_node = node.getchildren()[0]
+            try:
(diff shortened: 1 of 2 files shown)


bfontaine:
You should check the length of `node.getchildren()` instead. If it’s empty then no need to try to access `[0]` and catch an exception.

andrewserong:
@bfontaine thanks for the tip! I've updated the PR.

mvantellingen:
Thanks @andrewserong! Are you also able to add an unittest for this?

andrewserong:
Hi @mvantellingen - apologies for the delay, have added a unit test for this particular issue.

mvantellingen:
Thanks!!

PR #193, opened 2016-09-20
fix SOAP Header namespace for SOAP 1.2

vashek:
Also apparently relevant for issue 92 (ws-addressing)

### DIFF
--- a/src/zeep/wsdl/utils.py
+++ b/src/zeep/wsdl/utils.py
@@ -2,7 +2,7 @@ from lxml import etree
 
 
 def get_or_create_header(envelope):
-    header_qname = '{http://schemas.xmlsoap.org/soap/envelope/}Header'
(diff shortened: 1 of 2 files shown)


csinchok:
Would love to see this get merged, I've been running into this issue as well.

jeffrand:
Ditto what @csinchok said, I currently have a work around but I would much rather use this solution.

mvantellingen:
Thanks @vashek, so I'm going to do some work on this before I can merge it (unittest, and i'm not really happy with the import statement). I'll try to look at it this weekend.

vashek:
@mvantellingen Right, I'm not surprised. The import was there just for the assert anyway, so I removed both. And I added a unittest by basically copying and slightly modifying the one for SOAP 1.1. It does fail before the fix and succeed after it.

mvantellingen:
@vashek thanks again. really appreciated! 👍

Issue #187, opened 2016-09-16
Add support for concurrent requests via asyncio and aiohttp


chrisimcevoy:
I've been using zeep (which I've grown to love by the way!) to drive a PyQt application I'm developing which presents a UI to create/send/receive SOAP messages. 

In order to prevent blocking calls via the requests module suspending the UI, and for performance gains as end-users of my PyQt application might be making multiple calls simultaneously, I ended up subclassing several zeep classes to in essence port the package to asyncio and aiohttp.

Apologies if this has been discussed previously - I didn't see anything around this subject in the previous issues or PR's - but would you consider supporting this natively?

If there's interest in this I would be happy to put together a PR for peer review...

Many thanks.

mvantellingen:
Hi @chrisimcevoy, yeah I would definitely love to see support for asyncio! :-)

mvantellingen:
Closing for now since there is now basic support for asyncio. If there is interested in extending this then that can be done in separate issues.

Thanks again!!!

PR #207, opened 2016-10-03
Add support for concurrent requests via asyncio and aiohttp

chrisimcevoy:
As suggested in #187 

This is my first public PR, so please be gentle ;)

There are a significant amount of new classes here, because it's generally considered a bad idea to mix asynchronous and synchronous code. However, I've tried to keep the usage similar between the Client and the new AsyncClient where possible. I've also tried to inherit from existing classes to minimise impact, where it seemed logical to do so.

The major difference is that to instantiate the AsyncClient, you need to run AsyncClient.create() which is both a coroutine and a classmethod and returns the client instance. This my preferred way of working around the need to await coroutines in **init**() (see [](http://stackoverflow.com/a/33134213/5567657)). 

Once your AsyncClient is instantiated, you can call client.service.operation() as normal, but this is now a coroutine which needs to be awaited. I've included an example showing usage (examples/async_example.py).

There are however a few caveats of which I'm aware:
- I've used the async/await syntax which is only available from Python 3.5. This may not jive well with you as the rest of the project seems to strive for compatibility.
- Owing to my relative lack of experience with pytest, I haven't written any tests for this. Invite others with more experience to do so...?

### DIFF
--- a/examples/async_example.py
+++ b/examples/async_example.py
@@ -1,0 +1,48 @@
+import asyncio
+from lxml import etree
+from zeep.client import AsyncClient, AsyncTransport
+from zeep.exceptions import Fault
+
+
+"""Example using AsyncClient to make concurrent requests to the Global Weather service.
+Determines how many cities are on record for each country.
+Run this example multiple times (noting the order in which Countries appear in the console) to observe concurrency.
+"""
+
+def handle_fut(fut):
+    # Callback scheduled by add_done_callback - will be called when future is done.
+    for resp in fut.result():
+        if issubclass(resp.__class__, Fault):
+            print(resp)
+        else:
+            print(parse_response(resp))
+
+def parse_response(resp):
+    # Parse the response and return the number of countries on record for the country.
+    doc = etree.XML(resp)
+    country_name = doc.find('.//Country').text
+    no_cities = len(doc.findall('.//City'))
+    return '{} has {} cities on record.'.format(country_name, no_cities)
+    
+
+if __name__ == "__main__":
+
+    loop = asyncio.get_event_loop()
+    
+    # Pass asyncio event loop as first parameter to AsyncTransport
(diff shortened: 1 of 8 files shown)


mvantellingen:
Hi @chrisimcevoy,

First of all, this looks really cool! Thanks. So the issue now is that this uses 3.5+ only syntax, so we need to see if can we can make this work as something like a plugin that can optionally be enabled.  However to be honest, I first want to get a stable 1.0 release out before making these changes.

I will keep this PR and might build upon it for a later version.

Again really appreciated!

chrisimcevoy:
Absolutely - I imagine there would need to be some sort of difficult decision made in that regard.

You _could_ make this compatible with python 3.4 by instead coding the coroutines using the old `yield from`/`@asyncio.coroutine` decorator syntax, which _should_ also run ok when called from `async`/`await` code in python 3.5 and later as the 3.4 syntax is forwards-compatible.

Also, the 3.4 asyncio was [backported to 3.3](https://github.com/python/asyncio), it's just not in the standard libs for that version.

I've seen [trollius](https://github.com/haypo/trollius) discussed on other packages where python 2.7 support is needed, but unfortunately it's [no longer maintained](http://trollius.readthedocs.io/deprecated.html#deprecated). Sorry, I have no real expertise in that area as I pretty much exclusively work in 3.4 or later.

So, lots of different ways you could approach this down the line, each with their pros and cons. With that in mind, I completely understand your desire to reach 1.0 before dealing with all that!

mvantellingen:
Hi @chrisimcevoy, I've taken your code and incorporated in a way that makes is completely optional. See #251 .

Really cool!

mvantellingen:
Just merged the initial asyncio support. In the future I want to extend the support with the functionality in this PR. However I will be closing it for now, but this PR is referenced in the changelog!

Again many thanks!!

PR #209, opened 2016-10-03
Make etree.cleanup_namespaces call possible for lxml < 3.5.x

### LINKED ISSUE #197: Check lxml.etree version in cleanup_namespaces call
We are using lxml 3.3.5, because our system (Zato) has pinned that version of lxml.
However, the latest (since 0.15.0) zeep versions are not compatible anymore with this lxml version because of a call to etree.cleanup_namespaces (in wsa.py). The keyword argument top_nsmap has been added to cleanup_namespaces somewhere after lxml version 3.3.5

Is it possible to check for lxml version or do a try except and call cleanup_namespaces with the top_nsmap if needed? Or does that not make any sense?

For now we downgrade to zeep 0.14.0 to avoid this issue.

And, the lxml constraint needs to be updated (top_nsmap argument added in lxml 3.5.0b1).

jjmurre:
This fixes #197.

My first try was introspecting cleanup_namespaces to check the call signature for the 'top_nsmap' argument. However, it turned out that older (I tried 3.3.5) versions of lxml have a pure c implementation of this method that cannot be introspected, so I had to fall back to try-except.

I cannot oversee if just leaving out the top_nsmap argument does make any sense though.

### DIFF
--- a/src/zeep/wsa.py
+++ b/src/zeep/wsa.py
@@ -24,9 +24,12 @@ class WsAddressingPlugin(Plugin):
             WSA.To(binding_options['address']),
         ]
         header.extend(headers)
-        etree.cleanup_namespaces(
-            envelope, top_nsmap={
-                'wsa': 'http://www.w3.org/2005/08/addressing'
-            })
+        try:
+            etree.cleanup_namespaces(
+                envelope, top_nsmap={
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks! Merged it and decided to check the lxml version info instead of a try/except clause.

jjmurre:
Great!

PR #214, opened 2016-10-06
Adding an example of HTTP Authentication via the Transport class

samwyse:


mvantellingen:
Thanks you! Much appreciated 👍

PR #215, opened 2016-10-06
Make SOAP operation arguments available in Plugin.egress

ri8on:


mvantellingen:
Thanks! I think we want to encapsulate the args/kwargs or accept them as op_args / op_kwargs and not unpack them. Since now might error when the operation accepts a kwarg operation :-)

ri8on:
Yeah, you're right, it's my oversight )

mvantellingen:
Thanks. I'll need to add some code to make it backwards compatible but that should be doable. I'll look at it asap

ri8on:
Thanks!

vstoykov:
@ri8on your PR need to be rebased against master. Can you do that?

PR #230, opened 2016-10-23
Add a null check on the message result

javajenks:
Currently len() is done before checking whether the message result is None.


mvantellingen:
Thanks! I already had a commit ready to go for this, see d415fea24f2c1b105cba396ec437782ec7d9f3dc 

Let me know if this didn't solve your issue

javajenks:
oh cool, thats more elegant. thanks.

PR #235, opened 2016-10-23
Fixes for accepted_types variable on builtin types.

bjarnagin:
`six.string_types` returns a tuple of valid types. This tuple was being nested within another tuple during class instantiation of zeep builtins, which was causing `elements.py` to throw an `AttributeError` on line 153 during render:

```py
  File "/Users/billy/code/py/python-zeep/src/zeep/xsd/elements.py", line 153, in _render_value_item
    '%s.%s' % (t.__module__, t.__name__) for t in expected_types
  File "/Users/billy/code/py/python-zeep/src/zeep/xsd/elements.py", line 153, in <listcomp>
    '%s.%s' % (t.__module__, t.__name__) for t in expected_types
AttributeError: 'tuple' object has no attribute '__module__'
```

Where the listcomp in `type_names` was attempting to iterate through `expected_types`, which had the shape `(etree._Element, (str,))`.

This pull request initializes accepted_types to `six.string_types` when that is the only accepted type on the class, otherwise it is merged into the `accepted_types` tuple.

### DIFF
--- a/src/zeep/xsd/builtins.py
+++ b/src/zeep/xsd/builtins.py
@@ -87,7 +87,7 @@ class _BuiltinType(SimpleType):
 
 class String(_BuiltinType):
     _default_qname = xsd_ns('string')
-    accepted_types = (six.string_types,)
+    accepted_types = six.string_types
 
     def xmlvalue(self, value):
         return six.text_type(value if value is not None else '')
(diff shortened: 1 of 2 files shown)


mvantellingen:
Oh good one, didn't test that as good as i should.Thanks!!

PR #236, opened 2016-10-24
Fixes #232: Wrap children of an array in the appropriate element

### LINKED ISSUE #232: ArrayType serialization problem
_Using: zeep==0.19.0_

Either I'm not using the Array properly, or something goes wrong when serializing it. I expect an array to be serialized as: `

```xml
<values>
  <item>1</item>
  <item>2</item>
</values>
```
Instead, it's serialized as either the last element of a list, or (when nested) without the parent surrounded item element.

I added a test to the `test_wsdl_messages_rpc.py` to show the issue.

```python
def test_serialize_wsdl_array_of_simple_types():
    wsdl_content = StringIO("""
    <definitions xmlns="http://schemas.xmlsoap.org/wsdl/"
        targetNamespace="http://tests.python-zeep.org/tns"
        xmlns:tns="http://tests.python-zeep.org/tns"
        xmlns:impl="http://tests.python-zeep.org/tns"
        xmlns:xsd="http://www.w3.org/2001/XMLSchema"
        xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/"
        xmlns:soap="http://schemas.xmlsoap.org/wsdl/soap/"
        xmlns:soapenc="http://schemas.xmlsoap.org/soap/encoding/">
      <types>
        <schema xmlns="http://www.w3.org/2001/XMLSchema" targetNamespace="http://tests.python-zeep.org/tns">
          <complexType name="ArrayOfString">
            <complexContent>
              <restriction base="soapenc:Array">
                <attribute ref="soapenc:arrayType" wsdl:arrayType="xsd:string[]"/>
              </restriction>
            </complexContent>
          </complexType>
        </schema>
      </types>
      <portType name="SimpleTypeArrayPortType">
        <operation name="getSimpleArray">
          <input message="tns:getSimpleArrayRequest"/>
          <output message="tns:getSimpleArrayResponse"/>
        </operation>
      </portType>
      <binding name="SimpleTypeArrayBinding" type="tns:SimpleTypeArrayPortType">
        <soap:binding style="rpc" transport="http://schemas.xmlsoap.org/soap/http"/>
        <operation name="getSimpleArray">
          <soap:operation soapAction=""/>
          <input>
            <soap:body use="encoded" namespace="http://tests.python-zeep.org/tns" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/>
          </input>
          <output>
            <soap:body parts="return" use="encoded" namespace="http://tests.python-zeep.org/tns" encodingStyle="http://schemas.xmlsoap.org/soap/encoding/"/>
          </output>
        </operation>
      </binding>
      <message name="getSimpleArrayRequest">
        <part name="values" type="tns:ArrayOfString"/>
      </message>
      <message name="getSimpleArrayResponse"/>
    </definitions>
    """.strip())

    transport = DummyTransport()
    transport.bind(
        'http://schemas.xmlsoap.org/soap/encoding/',
        load_xml(io.open('tests/wsdl_files/soap-enc.xsd', 'r').read().encode('utf-8')))
    root = wsdl.Document(wsdl_content, transport)

    binding = root.bindings['{http://tests.python-zeep.org/tns}SimpleTypeArrayBinding']
    operation = binding.get('getSimpleArray')

    assert operation.input.signature() == 'values: ArrayOfString'

    ArrayOfString = root.types.get_type('ns0:ArrayOfString')
    val = ArrayOfString(['item', 'and', 'even', 'more', 'items'])

    serialized = operation.input.serialize(values=val)
    expected = """
        <?xml version="1.0"?>
        <soap-env:Envelope
            xmlns:ns0="http://tests.python-zeep.org/tns"
            xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">
          <soap-env:Body>
            <ns0:getSimpleArray>
              <values>
                <item>item</item>
                <item>and</item>
                <item>even</item>
                <item>more</item>
                <item>items</item>
              </values>
            </ns0:getSimpleArray>
          </soap-env:Body>
        </soap-env:Envelope>
    """
    assert_nodes_equal(expected, serialized.content)
```

Running the tests gave a few issues (see bottom), which led me to believe I am using the ArrayType incorrectly, but I don't see how else it should be used. 

The result is:

```bash
E       assert '<soap-env:En...v:Envelope>\n' == '<soap-env:Env...v:Envelope>\n'
E         Skipping 171 identical leading characters in diff, use -v to show
E         +   <values>items</values>
E         -   <values>
E         -         <item>item</item>
E         -         <item>and</item>
E         -         <item>even</item>
E         -         <item>more</item>
E         -         <item>items</item>
E         -       </values>
E               </ns0:getSimpleArray>
E             </soap-env:Body>
E           </soap-env:Envelope>
```

So, why is the values element not a list of item-elements, but just the string of the last item element?

**Sidenote: Problems when running the above tests:**
`AttributeError: 'String' object has no attribute '_value_class'`
This bug is also mentioned #226, but the restriction checking makes sense. I changed it to:

_xsd/elements.py:137_
```python
        if (
            self.restrict and not
            isinstance(value, (etree._Element, type(self.restrict)))

        ):
            raise TypeError((
                "Received object of type %r, " +
                "expected %s or etree.Element"
            ) % (type(value).__name__, type(self.restrict).__name__))
```

Then, the error would read: `TypeError: Received object of type 'str', expected String or etree.Element` which seems more correct but well, xsd.String and str probably result in the same thing.

For the test to continue, I simply removed that if-statement.

joeribekker:
I also changed an existing test that failed after the code changes, but the expected result looked incorrect to me. Hence, I changed that test as well.

### DIFF
--- a/src/zeep/xsd/elements.py
+++ b/src/zeep/xsd/elements.py
@@ -131,6 +131,11 @@ class Any(Base):
                     node = etree.SubElement(parent, 'item')
                     node.set(xsi_ns('type'), self.restrict.qname)
                     self._render_value_item(node, val)
+            elif self.restrict:
+                for val in value:
+                    node = etree.SubElement(parent, self.restrict.name)
+                    # node.set(xsi_ns('type'), self.restrict.qname)
+                    self._render_value_item(node, val)
             else:
                 for val in value:
                     self._render_value_item(parent, val)
--- a/tests/test_wsdl_arrays.py
+++ b/tests/test_wsdl_arrays.py
@@ -111,3 +111,60 @@ def test_complex_type():
         </document>
     """
     assert_nodes_equal(expected, node)
+
+
+def test_complex_type_without_name():
+    schema = xsd.Schema(load_xml("""
+    <xsd:schema
+        xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+        xmlns:wsdl="http://schemas.xmlsoap.org/wsdl/"
+        xmlns:SOAP-ENC="http://schemas.xmlsoap.org/soap/encoding/"
+        xmlns:tns="http://tests.python-zeep.org/tns"
+        targetNamespace="http://tests.python-zeep.org/tns">
+      <xsd:import namespace="http://schemas.xmlsoap.org/soap/encoding/"/>
+
+      <xsd:complexType name="ArrayObject">
+        <xsd:sequence>
+          <xsd:element name="attr_1" type="xsd:string"/>
+          <xsd:element name="attr_2" type="xsd:string"/>
+        </xsd:sequence>
+      </xsd:complexType>
+      <xsd:complexType name="ArrayOfObject">
+        <xsd:complexContent>
+          <xsd:restriction base="SOAP-ENC:Array">
+            <xsd:attribute ref="SOAP-ENC:arrayType" wsdl:arrayType="tns:ArrayObject[]"/>
+          </xsd:restriction>
+        </xsd:complexContent>
+      </xsd:complexType>
+    </xsd:schema>
+    """), transport=get_transport())
+
+    ArrayOfObject = schema.get_type('ns0:ArrayOfObject')
+    ArrayObject = schema.get_type('ns0:ArrayObject')
+
+    value = ArrayOfObject([
+        ArrayObject(attr_1='attr-1', attr_2='attr-2'),
+        ArrayObject(attr_1='attr-3', attr_2='attr-4'),
+        ArrayObject(attr_1='attr-5', attr_2='attr-6'),
+    ])
+
+    node = etree.Element('document')
+    ArrayOfObject.render(node, value)
+
+    expected = """
+        <document>
+            <ArrayObject>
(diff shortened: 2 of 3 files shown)


src/zeep/xsd/elements.py
```diff
@@ -131,6 +131,11 @@ def render(self, parent, value):
                     node = etree.SubElement(parent, 'item')
                     node.set(xsi_ns('type'), self.restrict.qname)
                     self._render_value_item(node, val)
+            elif self.restrict:
+                for val in value:
+                    node = etree.SubElement(parent, self.restrict.name)
```
mvantellingen: I think this should be self.restrict.qname ?

mvantellingen: Although this might depend on qualified vs unqualified elementForm state

mvantellingen:
Thanks!!

mvantellingen:
Seemed to work OK after testing so I've merged it. Thanks

PR #238, opened 2016-10-24
ListType should inherit from SimpleType

miketheman:
The list element has a parent element of SimpleType, not Type.

Signed-off-by: Mike Fiedler <miketheman@gmail.com>

I was uncertain on how to best provide a type-inheritance test for the list type.

### DIFF
--- a/src/zeep/xsd/types.py
+++ b/src/zeep/xsd/types.py
@@ -506,7 +506,7 @@ class ComplexType(Type):
         return value
 
 
-class ListType(Type):
+class ListType(SimpleType):
     """Space separated list of simpleType values"""
 
     def __init__(self, item_type):


mvantellingen:
Thanks, happy to accept the pr! I'm just wondering if this fixes a bug (and if so, should we add a unittest for it?)

miketheman:
@mvantellingen I was hitting a bug where I was getting a `NotImplementedError` when parsing a block that looks something like this:

```xml
<xsd:simpleType name="CountryNameType">
				<xsd:list>
					<xsd:simpleType>
						<xsd:restriction base="xsd:string">
							<xsd:enumeration value="None"/>
							<xsd:enumeration value="AlternateName"/>
							<xsd:enumeration value="City"/>
							<xsd:enumeration value="Code"/>
							<xsd:enumeration value="Country"/>
						</xsd:restriction>
					</xsd:simpleType>
				</xsd:list>
			</xsd:simpleType>
```

I couldn't figure out the right way to add this to a unit test - it's not a bug in the code per-se, only an inheritance of types - the doc line following the change reinforces the inheritance, but the subclassing was still pointing at `Type`.

mvantellingen:
See #241

Issue #229, opened 2016-10-22
Setting xsi:type attribute on "anyType" elements and adding xsi namespace to requests?


ginmrt:
Hello!

I am currently trying to generate the proper XML format for a specific "Any Type" element within my requests using Zeep.

The WSDL of this element is:

```
<xs:complexType name="Property">
  <xs:attribute name="Name" type="xs:string" use="required"/>
  <xs:sequence>
    <xs:element minOccurs="0" name="Value" type="xs:anyType"/>
  </xs:sequence>
</xs:complexType>
```

I have verified that the proper working format (in the case where the value is a string) for the resulting XML is something like:

```
<Property Name="details">
    <Value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:string">Some Text</Value>
</Property>
```

I am however unable to ensure the **xmlsn:xs** and **xsi:type** attributes are present in the XML contents that Zeep generates.

With many attempts, the closest I've been able to get was:

```
<Property Name="details">
    <Value>Some Text</Value>
</Property>
```

I've been playing around with AnyObject, ComplexType, AnyType and other elements of the Zeep library without success.

I am basically looking for pointers on how to properly add those two missing attributes to my Value element and how to properly add the "xsi" namespace to the SOAP envelope.

Zeep version: 0.18.1, installed via pip

Thank you.

mvantellingen:
Turns out that the anyType wasn't really supported well. I've just add a commit which allows you to do the following:
   
```python
from zeep import xsd

value = xsd.AnyObject(xsd.String(), 'Some Text')
obj = Property(Value=value)
```

I've updated the docs for this, see http://docs.python-zeep.org/en/master/datastructures.html#anytype-objects. 

Thanks for the detailed report btw, and let me know if you it still doesn't work

ginmrt:
Hello again, thanks for the quick response!

After quickly trying it out with the new commit and your example format on my end, this is the result I'm getting:

```
<Property Name="details">
    <Value xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:type="xsd:string">Some Text</Value>
</Property>
```

The service I'm sending requests to accepts **xsi:type="xs:string"**, whereas we now have **xsi:type="xsd:string"**.

I am not quite sure if "xs" and "xsd" are equivalent at this position or if "xs" is the only correct one.

If it helps, this is the error I am receiving, which leads me to think that maybe it should be "xs"? :  

```
[...]

PR #244, opened 2016-10-28
A test case to try to reproduce #229

pierredavidbelanger:
Just for fun, I tried to reproduce the problem (invalid XML where the xsd shorthand is never declared) mentioned in [this comment](https://github.com/mvantellingen/python-zeep/issues/229#issuecomment-255635448). I am one of the dev for the Java web service that is being called in the comment.

To my surprise, this test case is not failing. And I am not too sure if this is expected or not.
I am very noob with Python, so there may be plenty of things I overlooked.

I submit this PR anyways. Maybe you could have a use for it.

Great lib by the way!

### DIFF
--- a/tests/test_xsd_any.py
+++ b/tests/test_xsd_any.py
@@ -218,6 +218,63 @@ def test_element_any_type():
     assert item.something == 'bar'
 
 
+def test_array_of_element_any_type():
+    node = etree.fromstring("""
+        <?xml version="1.0"?>
+        <schema xmlns="http://www.w3.org/2001/XMLSchema"
+                xmlns:tns="http://tests.python-zeep.org/"
+                targetNamespace="http://tests.python-zeep.org/"
+                elementFormDefault="qualified">
+          <element name="container">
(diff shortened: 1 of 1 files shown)


tests/test_xsd_any.py
```diff
@@ -218,6 +218,63 @@ def test_element_any_type():
...
+    container_element = schema.get_element('{http://tests.python-zeep.org/}container')
+    array_of_property_type = schema.get_type('{http://tests.python-zeep.org/}array_of_property')
+    property_type = schema.get_type('{http://tests.python-zeep.org/}property')
+
+    property1 = property_type(name='a name', value=xsd.AnyObject(xsd.String(), 'a value'))
+    array_of_property = array_of_property_type(property=[property1])
+    container = container_element(properties=array_of_property)
+
+    node = etree.Element('document')
+    container_element.render(node, container)
```
mvantellingen: if you add etree.cleanup_namespaces(node) below this line then you will trigger the bug

Issue #260, opened 2016-11-16
Parsing WSDL returns TypeError: argument to reversed() must be a sequence 


kritsanapong:
Hi,

I am using zeep version 0.22.0 and was trying to parse a WSDL file, but got the error message "TypeError: argument to reversed() must be a sequence"

1. zeep version 0.22.0 on Python 3.4.4
2. WSDL location at https://webservices.sandbox.netsuite.com/wsdl/v2013_2_0/netsuite.wsdl
3. Example script
`
python -mzeep https://webservices.sandbox.netsuite.com/wsdl/v2013_2_0/netsuite.wsdl
`
4. Exception

```
Traceback (most recent call last):
   File "c:\Python34\Lib\runpy.py", line 170, in _run_module_as_main
      "__main__", mod_spec)
   File "c:\Python34\Lib\runpy.py", line 85, in _run_code
      exec(code, run_globals)
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\__main__.py", line 85, in <module>
      main(args)
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\__main__.py", line 74, in main
      client = Client(args.wsdl_file, transport=transport)
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\client.py",line 69, in __init__
      self.wsdl = Document(wsdl, self.transport)
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\wsdl\wsdl.py", line 61, in __init__
      root_definitions = Definition(self, document, self.location)
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\wsdl\wsdl.py", line 154, in __init__
      self.parse_types(doc)
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\wsdl\wsdl.py", line 268, in parse_types
      self.types.add_documents(schema_nodes, self.location)
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\xsd\schema.py", line 42, in add_documents
      document.resolve()
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\xsd\schema.py", line 275, in resolve
      schema.resolve()
   File "D:\src\workspace\env\partner_service\lib\site-packages\zeep\xsd\schema.py", line 287, in resolve
      _resolve_dict(self._types)
[...]

PR #261, opened 2016-11-16
test and fix for #260

jaceksnet:
if second extension in line adds a sequence to a type without a sequence there was this error


mvantellingen:
Hi @jaceksnet, thanks for the work (and adding the test!). I've cherry-picked them from your branch.

In the future could you create new branches for new functionality and submit pull-request for each branch separately? This PR currently contains unrelated changes. Anyway, I can always cherry-pick them but it's a bit more work ;-)

Thanks again

jaceksnet:
Yes, I know, sorry for that ;) Recently I was working with mercurial only and have bad habits for git ;)

PR #266, opened 2016-11-23
Remove Redundant Headers

dolimoon6000:
When creating a SOAP request and the request accepts multiple headers, if the user specifies only some of the headers in _soapheaders field, this library generates empty tags for all unspecified headers. These empty headers can sometimes cause problems with some clients. As a result a check is now made to remove headers that have no content.

### DIFF
--- a/src/zeep/wsdl/messages/soap.py
+++ b/src/zeep/wsdl/messages/soap.py
@@ -302,6 +302,10 @@ class SoapMessage(ConcreteMessage):
         else:
             raise ValueError("Invalid value given to _soapheaders")
 
+        # remove redundant headers
+        for x in header:
+            if len(x) == 0 and x.text is None:
+                x.getparent().remove(x)
         return header
 
     def _deserialize_headers(self, xmlelement):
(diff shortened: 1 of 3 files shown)


fmalina:
Please merge.

fmalina:
We actually need to exclude empty body tags too. Only include tags when we explicitly passed content to populate them, not just because the tag spec is in the WSDL. Does this make sense now?

mvantellingen:
So my initial feeling is that this is a workaround. I think that indeed we should stop sending all soap headers automatically if there is no value given. However the solution in that case is to not add the soap headers in the first place. Now we first add the headers and in a later step remove them again if they turn out to be empty. But I think we already know that they are empty when we add them.

fmalina:
Exactly! Just what we were discussing with Olu today.

I think it doesn't apply just to headers. Zeep objects with all default values set do get BIG.

mvantellingen:
Sorry for the slow response on this issue. I think i've added a solution which doesn't result in adding and then removing the xml node again.

Are you still able to give this a try?  (in master branch)

fmalina:
Olu? @dolimoon6000, @bartmalanczuk if you can test and this works, would you please provide feedback here and update python-netsuite requirements.txt to plain Zeep.

mvantellingen:
I'm closing this ticket for now as it should be fixed. If there are still any issues then please let me know so that I can re-open the issue.

olufusepump:
@mvantellingen, @fmalina, sorry for the late feedback. I can confirm that the issue has been resolved.

Issue #264, opened 2016-11-18
RuntimeError: maximum recursion depth exceeded


mmosca:
I get the above error when trying to load the attached wsdl with the zeep built from the master branch (ff9e5b1).

This wsdl was affected by issue 260, but the fix for 260 was not enough.

[wsdl.txt](https://github.com/mvantellingen/python-zeep/files/599992/wsdl.txt)

This is what the last few lines of the stack trace look like:

...
    parts.append('%s: %s' % (name, element.signature(depth)))
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\elements.py", line
367, in signature
    value = self.type.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\types.py", line 512
, in signature
    part = element.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\indicators.py", lin
e 196, in signature
    parts.append('%s: %s' % (name, element.signature(depth)))
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\elements.py", line
367, in signature
    value = self.type.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\types.py", line 512
, in signature
    part = element.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\indicators.py", lin
e 196, in signature
    parts.append('%s: %s' % (name, element.signature(depth)))
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\elements.py", line
367, in signature
    value = self.type.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\types.py", line 512
, in signature
    part = element.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\indicators.py", lin
e 196, in signature
    parts.append('%s: %s' % (name, element.signature(depth)))
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\elements.py", line
367, in signature
    value = self.type.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\types.py", line 512
, in signature
    part = element.signature(depth)
  File "c:\users\mbezerra4\hackaton\zeep-master\src\zeep\xsd\indicators.py", lin
[...]

PR #270, opened 2016-11-24
Fix for #264

jaceksnet:
I propose to fix this bug by using a tuple of element names as a depth parameter in signature method, instead of an integer. So not only we know how deep in recursion we are, but also for what elements signature was already created.


mvantellingen:
Thanks! Is there a specific reason to use a tuple() instead of a set()?

jaceksnet:
tuple() is immutable and easier to deal with in this situation, when used as a parameter, although I know searching in a set() is faster...

mvantellingen:
Okay! Merging this now as is, thanks again.

PR #278, opened 2016-12-15
asyncio: compatible with python 3.3 & 3.4

dzen:
I'm not sure it's something you want, but why not trying to be compatible
with python 3.4 ?

### DIFF
--- a/src/zeep/asyncio/bindings.py
+++ b/src/zeep/asyncio/bindings.py
@@ -1,3 +1,6 @@
+
+import asyncio
+
 from zeep.wsdl import bindings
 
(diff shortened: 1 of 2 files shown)


mvantellingen:
It looks good, thanks! Seems like a simple change (although I prefer the 3.5+ syntax).

I'll think about it (want to add unittests for the asyncio part first)

dzen:
Thank you for having a look on this !

Unfortunately, debian Jessie still rely on python 3.4. Debian stretch is programmed for first quarter this year, which will help a lot's of people to just use async / await :)

mvantellingen:
I've decided to make asyncio with zeep a 3.5+ only feature. Sorry. The PR was great though

Issue #281, opened 2016-12-16
Passing WSDL from string/buffer


nyov:
`Client` should accept a WSDL from a provided string or file handle instead of insisting on it's own loading routine.

While the standard `Transport` class can load from `file:///` URIs, sometimes a bit more flexibility and less magic would be useful.

mvantellingen:
Should already work via a filehandle (?)

Issue #277, opened 2016-12-12
Failing wsdl inspection with unknown transports


nyov:
`python -mzeep 'api/wsdl'` fails with
`NotImplementedError: Only soap/http is supported for now`.

Sadly this error message is rather useless without digging through the source code! It could at least say what it encountered instead ;)
In my case it could have said `Transport 'http://schemas.microsoft.com/soap/tcp' is unsupported`.

But mainly, this behavior seems counter to having multiple transport bindings, when the client fails on the first thing it _doesn't_ understand.

The WSDL (excerpt) in this case looks something like
```xml
<wsdl:binding name="BasicHttpBinding_IPublicInterfaceServiceContract" type="tns:IPublicInterfaceServiceContract">
	<soap:binding transport="http://schemas.xmlsoap.org/soap/http"/>
	<wsdl:operation name="abc"></wsdl:operation>
</wsdl:binding>

<wsdl:binding name="NetTcpBinding_IPublicInterfaceServiceContract" type="tns:IPublicInterfaceServiceContract">
	<wsp:PolicyReference URI="#NetTcpBinding_IPublicInterfaceServiceContract_policy"/>
	<soap12:binding transport="http://schemas.microsoft.com/soap/tcp"/>
	<wsdl:operation name="abc"></wsdl:operation>
</wsdl:binding>
```

While I'm not too knowledgeable in the world of SOAP, I think the client should just skip the unknown and continue with the transport that _does_ define `soap/http`, here.

As I now need to get this manually, I also lament the shortness of the "Non-default bindings" docs section. A code example there would be nice!

nyov:
Actually I just noticed I can't even get to the stage of using the other binding manually.
Just instancing `Client('api/wsdl')` is enough to throw the exception. Now this seems even more serious.
Looks like I have to chew and spoonfeed the WSDL to the client, for the time being.

mvantellingen:
> python -mzeep 'api/wsdl' fails with
> NotImplementedError: Only soap/http is supported for now.
> 
> Sadly this error message is rather useless without digging through the source code! It could at least say what it encountered instead ;)
> In my case it could have said Transport 'http://schemas.microsoft.com/soap/tcp' is unsupported.

Yes agreed, pull-requests are more then welcome :-)

> While I'm not too knowledgeable in the world of SOAP, I think the client should just skip the unknown and continue with the transport that does define soap/http, here.

Yup also agreed,  I will also accept happily accept a pull-request for this.

[...]

PR #282, opened 2016-12-16
Some patches

nyov:
It seems I forgot to create a PR with my changes... haha.

This is a collection of modifications I made. I couldn't figure out my core issue (#277) but I tried to work around this by passing a custom WSDL (#281).
For this I had to pass it in as data, so I added a way to do that. It's a bit complex though and not the nicest way, but it worked for me:
```py
from zeep import Client
from zeep.wsdl import Document
from zeep.transports import Transport

t = Transport()
wsdl = Document(open('My.wsdl', 'rb'), t)
client = Client(wsdl)
```
Perhaps I missed something there. But I hope it can be improved.

### DIFF
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -79,10 +79,12 @@ class Client(object):
     def __init__(self, wsdl, wsse=None, transport=None,
                  service_name=None, port_name=None, plugins=None):
         if not wsdl:
-            raise ValueError("No URL given for the wsdl")
+            raise ValueError("`wsdl` argument can not be empty")
 
-        self.transport = transport or Transport()
-        self.wsdl = Document(wsdl, self.transport)
+        if not isinstance(wsdl, Document):
+            wsdl = Document(wsdl, transport or Transport())
+        self.wsdl = wsdl
+        self.transport = self.wsdl.transport or Transport()
         self.wsse = wsse
         self.plugins = plugins if plugins is not None else []
 
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -14,7 +14,7 @@ class Transport(object):
     supports_async = False
 
     def __init__(self, cache=NotSet, timeout=300, operation_timeout=None,
-                 verify=True, http_auth=None):
+                 verify=True, http_auth=None, http_headers={}):
         """The transport object handles all communication to the SOAP server.
 
         :param cache: The cache object to be used to cache GET requests
@@ -24,6 +24,7 @@ class Transport(object):
         :param verify: Boolean to indicate if the SSL certificate needs to be
                        verified.
         :param http_auth: HTTP authentication, passed to requests.
+        :param http_headers: Custom HTTP headers dict, passed to requests.
 
         """
         self.cache = SqliteCache() if cache is NotSet else cache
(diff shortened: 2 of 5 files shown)


src/zeep/client.py
```diff
@@ -77,14 +77,16 @@ def __getitem__(self, key):
 class Client(object):
 
     def __init__(self, wsdl, wsse=None, transport=None,
-                 service_name=None, port_name=None, plugins=None):
+                 service_name=None, port_name=None, plugins=[]):
```
mvantellingen: This is actually pretty bad practice, see http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

src/zeep/transports.py
```diff
@@ -5,16 +5,19 @@
...
+ZEEP_USERAGENT = 'Zeep/%s (www.python-zeep.org)' % __version__
+
+
 class Transport(object):
     supports_async = False
 
-    def __init__(self, cache=NotSet, timeout=300, operation_timeout=None,
-                 verify=True, http_auth=None):
+    def __init__(self, cache=SqliteCache(), timeout=300, operation_timeout=None,
+                 verify=True, http_auth=None, http_headers={}):
```
mvantellingen: This will init the SqliteCache during startup/import. This isn't something we want since sqlite is not always available and we want users to let them choose other cache backends in that case.

src/zeep/transports.py
```diff
@@ -24,17 +27,18 @@ def __init__(self, cache=NotSet, timeout=300, operation_timeout=None,
...
         self.operation_timeout = operation_timeout
         self.logger = logging.getLogger(__name__)
 
         self.http_verify = verify
         self.http_auth = http_auth
-        self.http_headers = {
-            'User-Agent': 'Zeep/%s (www.python-zeep.org)' % (get_version())
+        self.http_headers = http_headers or {
+            'User-Agent': ZEEP_USERAGENT,
         }
```
mvantellingen: This isn't the same. In this case the user agent isn't set if someone passes custom headers

src/zeep/wsdl/bindings/soap.py
```diff
@@ -193,7 +193,8 @@ def parse(cls, definitions, xmlelement):
         soap_node = xmlelement.find('soap:binding', namespaces=cls.nsmap)
         transport = soap_node.get('transport')
         if transport != 'http://schemas.xmlsoap.org/soap/http':
-            raise NotImplementedError("Only soap/http is supported for now")
+            raise NotImplementedError("Transport '%s' is unsupported."
+                    " (Only soap/http is supported for now.)" % transport)
         default_style = soap_node.get('style', 'document')
```
mvantellingen: This one seems fine

src/zeep/transports.py
nyov: Hm, perhaps like with `NotSet`, pass the class-name instead of an object?
( `SqliteCache` instead of `SqliteCache()` )
Then instance the argument, if `callable()`?

src/zeep/client.py
nyov: Thanks. This is one thing I forget again and again over the ages. Just doesn't stick.

src/zeep/transports.py
nyov: I would actually just go with using a memory cache by default. No dependency issues, and I feel like a library shouldn't write files to disk by default. If this is in fact the case.
What do you think?

mvantellingen:
I'm going to close this one. If you are still interested in contributing new PR's are ofcourse welcome. But please create separate PR's for each change

nyov:
Thanks, but I'll pass.
I took the time to arrange the changes into self-contained commits.
Cherry-picking should have been easy enough.

Issue #254, opened 2016-11-08
TypeError raise on zeep.helpers.serialize_object when the object is a list of strings


smurfMT:
VERSION: zeep==0.18.1
WSDL: Not available
I have a element where one of its child elements is:
```
<element name="OutputAddress">
    <complexType>
        <sequence>
            <element maxOccurs="4" ref="xxx:AddressLine"/>
            <element ref="xxxPostCode"/>
        </sequence>
    </complexType>
</element>

<element name="AddressLine">
    <simpleType>
        <restriction base="string">
            <maxLength value="60"/>
            <minLength value="0"/>
        </restriction>
    </simpleType>
</element>
```

When calling `serialize_object()` on output address's parent it raises a type error when serialising `AddressLine` As the element is essentially a list of strings.

```
File "site-packages/zeep/helpers.py", line 16, in serialize_object
    value = obj[key]
TypeError: string indices must be integers
```

Proposed change: add another conditional to check if the object is a string and if it is return it.
```
if isinstance(obj, str):
    return obj
```

mvantellingen:
Should be fixed in master

PR #283, opened 2016-12-18
support string input in serialize_object - #254

daonb:


src/zeep/helpers.py
```diff
@@ -10,6 +10,9 @@ def serialize_object(obj):
     if obj is None:
         return obj
 
+    if isinstance(obj, basestring):
```
mvantellingen: six.text_type :-)

daonb:
Thanks @dshingarev . I've approved my PR based on the @mvantellingen comments and `six`

dshingarev:
Sounds good, thanks again.

mvantellingen:
I've just refactored the serializer so that this should work. Let me know otherwise

PR #285, opened 2016-12-20
Fix for WSSE UsernameToken not using the correct namespace

jdelasoie:
wsa.py may need some change too, but as I am not sure about this, I did not modify it.


mvantellingen:
Seems fine to me, thanks! Are you able to fix the tests so that i can merge it?

mvantellingen:
rebased and merged

PR #287, opened 2016-12-21
Enables custom HTTP headers to be passed to the Transport for each request.

sczk:
This pull request enables the user to specify custom HTTP headers in a dict when creating a Transport object. It would also enable the user to override the default user-agent if necessary. 

I use this in a project where the device requires a specific User-Agent and Cookie header in order to access the SOAP API.

I understand this may be a niche requirement, but thought I'd put up a pull request just incase anyone else thought it was a good idea for the project.

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -14,7 +14,7 @@ class Transport(object):
     supports_async = False
 
     def __init__(self, cache=NotSet, timeout=300, operation_timeout=None,
-                 verify=True, http_auth=None):
+                 verify=True, http_auth=None, http_headers={}):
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks! So this is basically the same issue as #296 

It's something I need to think about (regarding api design)

mvantellingen:
See #148 for a more in-depth comments regarding this

mvantellingen:
Master now accepts a requests.Session() instance so you can use that to add http headers.

See http://docs.python-zeep.org/en/master/transport.html#http-authentication

PR #296, opened 2016-12-28
Allow custom cert info

k15h0re:
Allow sending custom cert info to requests through Transport

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -14,7 +14,7 @@ class Transport(object):
     supports_async = False
 
     def __init__(self, cache=NotSet, timeout=300, operation_timeout=None,
-                 verify=True, http_auth=None):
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks for the PR.

So a lot of times I receive pr's to add a specific kwarg for requests. I've decided that supporting all kwargs which are support by requests doesn't really result in a nice api. 

I think subclasses the transport class and then altering create_session() seems good, or perhaps just transport.request.cert = x

I need to think about this some more ;-)

mvantellingen:
See #148 for a more in-depth comments regarding this

k15h0re:
Sure. I did this for a use case I had at work. And I saw a few issues on
the same where the workaround was to set verify to false. So I thought this
would be a good feature to have.
Thanks for taking a look at it. Cheers.

PR #302, opened 2017-01-08
Multipart response support

dwapstra:
Added support for multipart responses for SOAP attachments. This works for file attachments. If a multipart is text/xml, it seems the xml processing is not working correctly for some reason.

The returned object is a list of parts from the multipart decoder.

### DIFF
--- a/docs/multipart.rst
+++ b/docs/multipart.rst
@@ -1,0 +1,25 @@
+======================================
+Multipart support for SOAP Attachments
+======================================
+
+If the server responds with a Content-type: multipart, the code will process
+the parts individually and return a list of objects. Typically this will be
+a ``BodyPart`` object from request_toolbelt.
+
+The ``BodyPart`` object is a ``Response``-like object to an individual
+subpart of a multipart response.
+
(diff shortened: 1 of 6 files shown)


mvantellingen:
First impression is that it looks really good, thanks!

The only issue for now is that this doesn't work with the asyncio part. It would be cool if this code was transport agnostic (e.g. works with requests as well as asyncio). But that seems doable, let me know if you want to take that up or otherwise I can do it myself.

Thanks again.

dwapstra:
I am not familiar with asyncio, if you can take that, it would probably be better. I can look into if if you have some pointers where to start. I did not see any test cases related to asyncio so not sure how to approach this.

mvantellingen:
hi @dwapstra, 

I've implemented the support and re-used your code mostly. See https://github.com/mvantellingen/python-zeep/commit/57f6711307e8e2e150cfa2c2fef874cf23fff668

Thanks again, and let me know if you have some remarks!

mvantellingen:
I'm closing this ticket for now as it should be fixed. If there are still any issues then please let me know so that I can re-open the issue.

Issue #123, opened 2016-07-11
serializing objects with several levels does not work


YAmikep:
Hello

Since I upgraded to 0.12, I am having issues serializing the zeep objects into python types.
It seems like it's not working when there are several levels: the deeper levels are not serialized.

Do the dynamic zeep objects inherit from a common class like `zeep.objects.Object`?
That would allow to use `isinstance(obj, zeep.objects.Object)` when going through the data and transform the zeep objects recursively?

Thanks

mvantellingen:
Not entirely sure what you mean. Is the XML send to the server invalid or are is the helpers.serialize_object() broken?

YAmikep:
I think `helpers.serialize_object()` is broken as it does not serialize all the inner objects.

For example:

```
# The object returned by zeep
result = {
    'AAA': [
        {
            'ID': 111,
            'BBB': {
                'CCC': [
                    {
                        'FromDate': datetime.datetime(2016, 7, 3, 0, 0),
                        'ToDate': datetime.datetime(2016, 7, 12, 0, 0),
                        'Desc': 'bla',
                    },
                    {
                        'FromDate': datetime.datetime(2016, 7, 13, 0, 0),
                        'ToDate': datetime.datetime(2016, 7, 20, 0, 0),
                        'Desc': 'hello',
                    },
					...
                ]
            }
        },
		...
	]
}

type(result)
zeep.objects.ArrayOfAAA

result2 = zeep.helpers.serialize_object(result)

type(result2)
collections.OrderedDict

# After serializing the object, the inner objects are still zeep objects
type(result2['AAA'][0])
Out[15]: zeep.objects.AAA

type(result2['AAA'][0]['BBB'])
zeep.objects.ArrayOfCCC

type(result2['AAA'][0]['BBB']['CCC'][0])
zeep.objects.CCC

```

As a workaround, if all the dynamic classes would inherit from a common `zeep.objects.Object` class, it would be easier for me to go through the data and convert them manually.

Thanks

mvantellingen:
Ok thanks, will fix it asap. All objects actually do extend from zeep.xsd.valueobjects.CompoundValue

mvantellingen:
Should be fixed in master, if not then let me know

YAmikep:
Hum I don't think it completely fixes the issue.
My object is:
```
> type(o)
zeep.objects.ArrayOfCompanies
> o
{
    'Companies': [
        {
            'ID': 111,
            'Name': None
        },
[...]

PR #306, opened 2017-01-12
Fix helpers.serialize_object

YAmikep:
- Fix helpers.serialize_object for lists of ints (#123).
 - Allow to change the dict type used in helpers.serialize_object.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,7 @@
+ - Fix helpers.serialize_object for lists of ints (#123).
+ - Allow to change the dict type used in helpers.serialize_object.
+ 
+ 
 0.24.0 (2016-12-16)
 -------------------
  - Don't fail the parsing of responses if an xsi:type references an non-existing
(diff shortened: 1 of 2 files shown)


mvantellingen:
I've just refactored the helpers.serialize_object() and now this should work.

YAmikep:
@mvantellingen You did not want to keep the `dict_type` parameter? It would be handy.

mvantellingen:
Maybe, what is exactly the use-case for it?

YAmikep:
The type of dict returned is currently hard coded in the function: `OrderedDict`.
I'd like to get simple dicts instead.
Having a parameter would at least give the option to change the type of object to use.

mvantellingen:
I understand, and although the change is pretty easy i'm more interested in the idea behind it. Is it required for json or something else?

YAmikep:
Yes, I store the output as json later on and was having issues with an OrderedDict.
Also, an OrderedDict is less efficient and if we do not need to keep the order, we might as well use a simpler dict.
Anyways, I think having at least the option is better as the user can make his own decision regarding the type to use.

mvantellingen:
Okay fine ;-)

It's in master

YAmikep:
Cool thanks. 
FYI, I added a comment in the commit: you need to pass the parameter to the function in the recursive calls as wel..

PR #332, opened 2017-02-02
Add xml_huge_tree option to client, to allow very large XML responses.

TWAC:
I have looked into switching from suds-jurko to zeep. One issue I came across was with very large responses.

In older versions (before 9ff52c8e1?) they where silently truncated. Now they give
```zeep.exceptions.XMLSyntaxError: Invalid XML content received (xmlSAX2Characters: huge text node, line 4, column 5000144)```

I found the reason to be security features in libxml2, this PR adds an option to disable them.

### DIFF
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -108,12 +108,15 @@ class Client(object):
                       first port defined in the service element in the WSDL
                       document.
     :param plugins: a list of Plugin instances
+    :param xml_huge_tree: disable lxml/libxml2 security restrictions and
+                          support very deep trees and very long text content
 
 
     """
 
     def __init__(self, wsdl, wsse=None, transport=None,
-                 service_name=None, port_name=None, plugins=None, strict=True):
+                 service_name=None, port_name=None, plugins=None,
+                 strict=True, xml_huge_tree=False):
         if not wsdl:
             raise ValueError("No URL given for the wsdl")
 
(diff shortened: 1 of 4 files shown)


mvantellingen:
Thanks! It looks good but I need to think about the external API. I think the flag xml_huge_tree to the Client() is a bit too much magic

TWAC:
Sure, tell me if I should move it somewhere else.

TWAC:
Any decision made yet?

mvantellingen:
I'll look at it asap. Getting a bit behind :-)

TWAC:
By the way, the default in 1.4.1 is non strict parsing, so it is back to silently truncating data.
[def parse_xml(content, transport, base_url=None, **strict=False**):](https://github.com/mvantellingen/python-zeep/blob/1.4.1/src/zeep/loader.py#L21)

This also contradicts the documentation.
[By default zeep will operate in 'strict' mode.](https://github.com/mvantellingen/python-zeep/blob/1.4.1/docs/client.rst#strict-mode)

mvantellingen:
Oh good find, yeah it should be strict by default, i'll look at it

suchanun:
I need to deal with large xml response, after setting `xml_huge_tree=True, raw_response=True`, still get this error:

`Server returned response (200) with invalid XML: Invalid XML content received (xmlSAX2Characters: huge text node, line 1, column 11108910)
...
<SuppMsgOut xsi:type="xsd:string">Exception of type \\\'System.OutOfMemoryException\\\' was thrown.</SuppMsgOut>

`

Is there any workaround possible?

PR #341, opened 2017-02-10
default to utf-8 for responses without encoding

mtrbean:


mvantellingen:
Thanks! Are you able to add a unittest?

mtrbean:
Hi @mvantellingen I just added a test for the change. Thanks!

PR #344, opened 2017-02-13
Fix exception referenced before assignment when decode SOAP Multipart

Falldog:
If there is only one part in multipart result, it will cause exception 
`UnboundLocalError: local variable 'message_pack' referenced before assignment`

Move `message_pack` for initialization


mvantellingen:
👍

PR #347, opened 2017-02-15
set default encoding utf-8 for wsdl attachments headers

Falldog:
Fix the issue if `part.encoding` is None will cause header string decode() exception

should get default encoding as `utf-8` if `part.encoding` is None

### DIFF
--- a/src/zeep/wsdl/attachments.py
+++ b/src/zeep/wsdl/attachments.py
@@ -49,8 +49,9 @@ class MessagePack(object):
 
 class Attachment(object):
     def __init__(self, part):
+        encoding = part.encoding or 'utf-8'
         self.headers = CaseInsensitiveDict({
-            k.decode(part.encoding): v.decode(part.encoding)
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks! Are you able to add a unittest for this? otherwise i can take care of that when I have some time available

Falldog:
I had browse test cases, there is no test case for multipart response.
I'm not sure how to write the test case in python-zeep style.
Maybe you could give me some hint to guide me how to deal with multipart test-case.

dwapstra:
@Falldog there is a test case for multipart here https://github.com/mvantellingen/python-zeep/blob/master/tests/test_wsdl_soap.py#L146

Falldog:
@mvantellingen, I think Pull Request #341 testing case & solution will cover the changed of this commit.

And it will assign `decoder` a default encoding `utf-8`, Attachment will never get `None` from `part.encoding`

PR #351, opened 2017-02-16
Fix for SOAP resonse ending of '--' with requests_toolbelt 0.7.1

dwapstra:
Encountered a SOAP response that did not end with --\r\n, this is fixed in the latest requests_toolbelt


mvantellingen:
thanks!

Issue #358, opened 2017-02-23
absolute_location does not support base urls with file as scheme


aerandir92:
Version 1.1.0

When you send a location, say 'TestWSDL.xsd', with base 'file:///tmp/scheme', the result ends up being something like /home/usr/file:/tmp/scheme/testWSDL.xsd
Simply adding file to the schemes in absolute_location should fix this. 
I can create a pull-request when I get home

Console output showing issue:
https://gist.github.com/aerandir92/e62d4bc51d75c6444edf53c01b902824

mvantellingen:
Merged your pr, thanks

PR #359, opened 2017-02-23
Added file to schemes in absolute_location

aerandir92:
Should fix merging of paths when base already contains an absolute location

Fixes issue #358


mvantellingen:
Thanks

PR #366, opened 2017-02-27
fixed bug with incorrectly looking up types

brouwert:


brouwert:
I see the `test_attribute_union_type`  test case fails, let me look into that.

src/zeep/xsd/visitor.py
```diff
@@ -802,10 +802,11 @@ def visit_union(self, node, parent):
         """
         # TODO
         members = node.get('memberTypes')
+        default_namespace = node.nsmap.get(None)
         types = []
         if members:
             for member in members.split():
-                qname = as_qname(member, node.nsmap, self.document._target_namespace)
+                qname = as_qname(member, node.nsmap, default_namespace or self.document._target_namespace)
```
mvantellingen: This should already be handled in the as_qname() function

brouwert: Take this example:

```
<?xml version="1.0"?>
<schema
    xmlns="http://www.w3.org/2001/XMLSchema"
    xmlns:tns="http://tests.python-zeep.org/"
    targetNamespace="http://tests.python-zeep.org/"
    elementFormDefault="qualified">
  <element name="item" type="tns:Date"/>
  <simpleType name="Date">
    <union memberTypes="date tns:MMYY"/>
  </simpleType>
  <simpleType name="MMYY">
    <restriction base="string" />
  </simpleType>
</schema>
```

Looking up type `date` for the `Date` simpleType will fail because it will try to look up the type in the targetNamespace. My understanding is that it *should* be looked up in the default namespace because it doesn't have a prefix, and the default namespace exists. 

Only when *defining* a type it should be defined in the targetNamespace, in this case its about referring to a type.

I encoutered this problem while trying to use the zeep library with the Kadaster, Dutch government service.

mvantellingen: Ah yeah, you're right. Then leaving the self.document._target_namespace as arg should fix it perhaps. I can look at it later today

brouwert: Btw, when loading the XSD XML from the union testcases in Altova XMLSpy I get validations errors, some types cannot be resolved. I don't know who's in the wrong here, Altova or the XSD in the testcase ...

Issue #383, opened 2017-03-17
Zeep client are not closes created sessions


penguinolog:
zeep 1.3.0 are not closes created requests.Session/aiohhtp.Session in destructor.
This behavior is correct for user-constructed session instances, but incorrect for built in constructor.
As result, aiohttp.Session produces warnings in destructor.

Solution: mark session creation status in `__init__` and handle it in `__del__`.

PR #384, opened 2017-03-17
Fix: Zeep client are not closes created sessions

penguinolog:
Close aiohttp.Session in destructor, if constructed during Transport instance creation.
Closes-bug: #383


mvantellingen:
Thanks!

PR #386, opened 2017-03-17
Fix xsd:boolean defaults

bartekwojcicki:
Passing "false" or "0" to xsd:boolean default value resulted in
returning "true".


tests/test_xsd_integration.py
```diff
@@ -476,6 +476,44 @@ def test_defaults():
...
+            </xsd:complexType>
+          </xsd:element>
+        </xsd:schema>
+    """.strip())
+
+    schema = xsd.Schema(node)
+    container_type = schema.get_element(
+        '{http://tests.python-zeep.org/}container')
+    obj = container_type()
+    assert obj.foo == "false"
```
bartekwojcicki: Not sure what should be expected here:
```
assert obj.foo == "false"
assert obj.bar == "0"
```
or
```
assert obj.foo is False
assert obj.bar is False
```
To achieve the latter, types' `pythonvalue` should be called somewhere (probably in `Element.default_value` property?).

lovmat:
+1

mvantellingen:
Rebased and merged, thanks!

PR #394, opened 2017-03-21
Improve docs cross-references

singingwolfboy:
See http://www.sphinx-doc.org/en/stable/domains.html#the-python-domain for more info.


mvantellingen:
Thanks!!! 👍

PR #406, opened 2017-03-28
Test input parameter against None in client, fixes #404

### LINKED ISSUE #404: Unexpected behavior in `Client` constructor when `transport` is a falsey object
The `Client` constructor checks the `transport` argument as follows: `self.transport = transport or Transport()`. This means that if a falsey object is passed as the `transport` argument, a default transport is silently created and used instead of causing an error. This should not happen in normal use, but a better check should be able to point out unexpected bugs.

I've discovered this behavior while debugging this piece of code:

```python
transport = Transport().get(...) # this is wrong
# transport is now a Response(400) object, which is falsey
client = Client(transport=transport)
# client is created with a default Transport object
# and behaves as normal
```

Changing the check to `transport if transport is not None else Transport()` would have at least accepted the (invalid) `Response` object and caused an `AttributeError` quickly. The current behavior led to a weird semi-working state despite the obviously incorrect call arguments.

P.S. Sorry for the stream of consciousness presentation, I just want to make things a bit more robust.

nanonyme:


mvantellingen:
Thanks!

PR #414, opened 2017-04-06
json compatible output printing

dwapstra:
Small changes to output printing makes it easier to load the resulting data with JSON parsers

### DIFF
--- a/src/zeep/xsd/printer.py
+++ b/src/zeep/xsd/printer.py
@@ -28,7 +28,7 @@ class PrettyPrinter(object):
             if num > 0:
                 for i, (key, value) in enumerate(obj.items()):
                     write(' ' * (indent * level))
-                    write("'%s'" % key)
+                    write('"%s"' % key)
                     write(': ')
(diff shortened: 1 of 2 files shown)


src/zeep/xsd/printer.py
```diff
@@ -53,7 +53,7 @@ def _format(self, obj, stream, indent=4, level=1):
                 write(' ' * (indent * (level - 1)))
             write(']')
         else:
-            value = repr(obj)
+            value = '%s' % obj
```
mvantellingen: This change will return __str__ instead of __repr__. Is that correct?

src/zeep/xsd/printer.py
```diff
@@ -64,4 +64,4 @@ def _format(self, obj, stream, indent=4, level=1):
                     if i < num - 1:
                         write('\n')
             else:
-                write(value)
+                write('"%s"' % value)
```
mvantellingen: Are double quotes escaped properly here?

Issue #415, opened 2017-04-07
aioresponses in setup requirements while used for tests only


penguinolog:
`aioresponses` package is in setup requirements while it's not needed for package functionality.

`FPM` package builder could not build correctly package with zeep[async] due to aioresponses requirement and aioresponses package internals.

Solution for zeep: drop aioresponses requirement from setup.py and use in tox.ini only.

PR #419, opened 2017-04-13
Fix: aioresponses in setup requirements while used for tests only #415

penguinolog:
Drop aioresponses from setup requirements


mvantellingen:
Thanks, fixed it slightly different in 1.5

PR #430, opened 2017-05-03
Fix error caused by storing _xsd_type: None during unpickling

sankalpanand:


sankalpanand:
My PR fails `test_pickle()` with this error. @mvantellingen can you tell what might be missing here?
`pickle.PicklingError: Can't pickle <class 'zeep.objects.ComplexType'>: it's not found as zeep.objects.ComplexType`

PR #433, opened 2017-05-09
Adding support for elements which use a substitutionGroup

ashleysommer:
First attempt at adding support for elements which use a `substitutionGroup`.

This allows an any element to be substituted for any element which is within the first element's `substitutionGroup`.

This also overrides the keyname of the element in the `kwargs` struct, so that an substituted element may be referred to by its correct name.

This is tested and working on several of my trivial local tests. No real tests written yet.


ashleysommer:
Added a test to cover this new functionality

ashleysommer:
Added more tests to (attempt to) increase coverage.

ashleysommer:
Rebased all commits in feature-branch to master.

ashleysommer:
Added some more tests on the new `element_parse_substituted()` function directly, to increase code coverage by exploiting edge cases.

ashleysommer:
Finally passed tests! Any chance of getting this merged? Or suggestions for what else is required?

mvantellingen:
Thanks for this! My first though was that this makes a number of methods considerably more complex and hard to follow. So perhaps we need to do some more abstraction / move code to separate methods.

I'll take a more in depth look at it asap

ashleysommer:
I agree. It is probably not the ideal way to integrate such a feature.
However given I don't know much about the architecture of Zeep, I feel I did the best I could.
If there is a better way to implement it, I am happy to comply.

In its current form, I am using Zeep patched with this feature in my codebase, and it fulfils my requirement.

Nuytho:
Hello,
I'd like to ask, Is there any work going on substitutionGroup support? I would really apreciate any info about this

thank for all the good work and hopefully for a fast asnswer

ashleysommer:
@Nuytho You can use my fork here: https://github.com/ashleysommer/python-zeep
That has Substitution Group support. But but beware, it is based on zeep v2.2.0 and has not been updated in about two years.

Nuytho:
@ashleysommer  thank you very much, but that kinda raised another kinds of errors that were fixed in later versions of zeep 😞  seems like this webservice is unhadleable in python in any normal way I will have to figure out some workaround

PR #434, opened 2017-05-09
Wrong base url in nested includes

jhorman:
Nested includes currently use the original document as the base url, leading to include failures.

document 1

```
<xs:include schemaLocation="chameleon/something.xsd"/>
```

document 2

```
<xs:include schemaLocation="something_else.xsd"/>
```

In the example above `something_else.xsd` will fail to load. With this change, the base url for document 2 is temporarily changed `http://.../chameleon/something.xsd`.


mvantellingen:
Looks good. Thanks!

PR #450, opened 2017-05-26
fix automatic "nonce" generation

lucasfcardozo:
The "nonce" was not being encoded (for UTF-8) when automatically generated


mvantellingen:
hi @lucafcardozo, thanks for the PR. Unfortunately the tests are failing. Do you want to take a look at that? Thanks

PR #457, opened 2017-06-01
Don't expect unicode response data

dmollerm:
Create 2 tests that assert a TransportError is raised if the content can be parsed as xml, regardless of the character encoding.

Don't try to encode the webservice response content as unicode, as this may fail.


mvantellingen:
thanks

PR #459, opened 2017-06-06
fix Remove wrapping object during deserialization of SOAP:Envelope

lancondrej:
Remove wrapping object during deserialization of SOAP:Envelope only if it hasn't got any attributes.


mvantellingen:
Thanks, merged. Are you also able to add a unittest for this?

lancondrej:
Yes, sure. I'll add a test ASAP.

Issue #454, opened 2017-05-30
RecursionError (infinite recursion while parsing a WSDL file)


nagylzs:
Hi!

I get a RecursionError (infinite recursion) when trying to load a WSDL document.

All files needed to reproduce are here: 

https://gist.github.com/nagylzs/aa150a696d0db16c1af3cf70f14448da

To reproduce the error, run this:

python3.6 -m zeep sts-username.wsdl 

Python version:

Python 3.6.1 (default, Mar 22 2017, 10:05:53) [GCC 5.4.1 20170304] on linux

Modules installed:

appdirs (1.4.3)
asn1crypto (0.22.0)
cached-property (1.3.0)
certifi (2017.4.17)
cffi (1.10.0)
chardet (3.0.3)
cryptography (1.8.1)
defusedxml (0.5.0)
idna (2.5)
isodate (0.5.4)
lxml (3.7.3)
ndg-httpsclient (0.4.2)
packaging (16.8)
pip (9.0.1)
pkgconfig (1.2.2)
pyasn1 (0.2.3)
pycparser (2.17)
pycurl (7.43.0)
pygobject (3.20.0)
pyOpenSSL (17.0.0)
pyparsing (2.2.0)
python-apt (1.1.0b1)
pytz (2017.2)
requests (2.17.3)
requests-toolbelt (0.8.0)
setuptools (35.0.2)
six (1.10.0)
ssh-import-id (5.5)
tornado (4.5.1)
unattended-upgrades (0.1)
urllib3 (1.21.1)
wheel (0.29.0)
xmlsec (1.0.9)
zeep (2.0.0)

nagylzs:
This and #451 might be the same.

mvantellingen:
This doesn't contain a runnable example, but might indeed be the same as #451.

If it isn't fixed in master could you _please_ create a runnable example?

nagylzs:
This is runnable, as stated: "python3.6 -m zeep some_wsdl_with_recursive_definitions.wsdl". (In other words:  the program is this: "python3.6 zeep.py some_recursive_data_structure.wsdl") I have narrowed down the problem to the dump() method, and inside that the various descendants of the Base.signature() method. Looks like the wsdl parsing is fine. The problem is that dump() calls signature in an infinite recursion for any recursive data structure.

[...]

Issue #451, opened 2017-05-30
zeep 2.0.0 with issue RuntimeError: maximum recursion depth exceeded


xudesheng:
1) Zeep version: 2.0.0
2) WSDL: https://gist.github.com/xudesheng/891eab5324135aeb00b86bd72d7ef343
3) Example Script:

from zeep import Client
client = Client('http://demo:6666/ProcessDefinitions/DataProcessing/SOAPServer?wsdl')
client.wsdl.dump()

RuntimeError: maximum recursion depth exceeded

Thanks for taking the time to report an issue!

Since Zeep is a module I've created and currently maintain mostly in my spare
time I need as much information as possible to quickly analyze/fix issues.

Please provide the following information:

1. The version of zeep (or if you are running master the commit hash/date)
2. The WSDL you are using
3. An example script

Also please don't paste WSDL/XSD file contents directly into the issue since 
that makes it hard to read, instead please create a gist at 
https://gist.github.com and link that here.

Please see http://docs.python-zeep.org/en/master/reporting_bugs.html for 
more information.

mvantellingen:
Hi, thanks for the info, can you give master a try? It should work there

PR #462, opened 2017-06-09
Signature/dump works with recursive data structures

nagylzs:
This should solve issues #454 and #451. The main change is that signature() methods are keeping track the path of the object, and it will not dump nested elements if they are already in the path.

### DIFF
--- a/src/zeep/xsd/types/complex.py
+++ b/src/zeep/xsd/types/complex.py
@@ -5,6 +5,7 @@ from itertools import chain
 
 from cached_property import threaded_cached_property
 
+from zeep import utils
 from zeep.exceptions import UnexpectedElementError, XMLParseError
 from zeep.xsd.const import NotSet, SkipValue, xsi_ns
 from zeep.xsd.elements import (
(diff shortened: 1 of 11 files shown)


src/zeep/xsd/types/complex.py
```diff
@@ -431,14 +432,20 @@ def restrict(self, base):
...
-    def signature(self, schema=None, standalone=True):
+    def signature(self, schema=None, standalone=True, path=None):
         parts = []
+
+        path = utils.extend_path(path, self)
+
         for name, element in self.elements_nested:
-            part = element.signature(schema, standalone=False)
+            if element in path:
+                part = '%s(...)' % name
```
nagylzs: Using a list is not effective. We could use a set instead of the list. Sadly, path elements are not hashable.

mvantellingen:
I had something like that but removed it again since I didn't really like it, see 965a5acf357c08716db079feb63d9a52a1857430

Is this actually still an issue ? Since i mentioned in the issue that i've fixed it right?

mvantellingen:
Closing this one since this the original issue was already resolved

PR #469, opened 2017-06-14
Fix parsing sequence with multiple annotations

antonech:
Visitor fails when sequence have multiple annotations.

example:
```
         <xsd:sequence>
            <xsd:annotation>
                <xsd:documentation>Data type of this variable</xsd:documentation>
            </xsd:annotation>
            <xsd:element name="dataType" type="tlpa:VariableType" />
            <xsd:annotation>
                <xsd:documentation>Display type of this variable</xsd:documentation>
            </xsd:annotation>
            <xsd:element name="displayType" type="tlpa:VariableDisplayType" />
           .
           .
           .
        </xsd:sequence>
```


mvantellingen:
I don't think this is actually valid. The xml spec define xsd:sequence like ``(annotation?,(element|group|choice|sequence|any)*)`` which means that there is at most one optional annotation.

Do you have an example xsd?

antonech:
Thank you for you help. You are right it is not valid case, i have managed to fix xsd file itself.

PR #325, opened 2017-01-29
support XML-binary Optimized Packaging (XOP) response

mvantellingen:
Converted work done by @vashek  to a PR to remind me to merge this somewhere post 1.0

PR #475, opened 2017-06-18
Add support for XOP messages

mvantellingen:
This is rather naive implementation but it should work nevertheless

Previous work on this was done by Vashek (thanks!) see #325

### DIFF
--- a/src/zeep/wsdl/attachments.py
+++ b/src/zeep/wsdl/attachments.py
@@ -75,6 +75,6 @@ class Attachment(object):
         if encoding == 'base64':
             return base64.b64decode(content)
         elif encoding == 'binary':
-            return content
+            return content.strip(b'\r\n')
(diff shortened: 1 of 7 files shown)


src/zeep/wsdl/attachments.py
```diff
@@ -75,6 +75,6 @@ def content(self):
         if encoding == 'base64':
             return base64.b64decode(content)
         elif encoding == 'binary':
-            return content
+            return content.strip(b'\r\n')
```
vashek: Why the strip? Looks strange to me, and potentially wrong, what if the binary content happens to have those bytes at the start and/or end on purpose?

lazka: see https://github.com/mvantellingen/python-zeep/issues/1481

Issue #479, opened 2017-06-23
Duplicate Body within SOAP request message


dacut:
This is happening on both Python 2.7 and Python 3.6. Zeep version is 2.2.0 (`zeep-2.2.0-py2.py3-none-any.whl`), installed via pip.

This is from a NOAA service whose WSDL lists an incorrect endpoint, so I've patched the WSDL here: https://tidal-harmonics.s3-us-west-2.amazonaws.com/ActiveStationsFixed.wsdl

Sample script that fails:

    import zeep
    active_stations = zeep.Client("https://tidal-harmonics.s3-us-west-2.amazonaws.com/ActiveStationsFixed.wsdl")
    active_stations.service.getActiveStationsV2()

Results in:

    zeep.exceptions.Fault: No such operation 'Body'

This is an issue when creating the request:

    >>> env = active_stations.create_message(active_stations.service, "getActiveStationsV2")
    <Element {http://schemas.xmlsoap.org/soap/envelope/}Envelope at 0x10a90de60>
    >>> env.getchildren()
    [<Element {http://schemas.xmlsoap.org/soap/envelope/}Body at 0x10a90d998>]
    >>> env.getchildren()[0].getchildren() # Should be empty here
    [<Element {http://schemas.xmlsoap.org/soap/envelope/}Body at 0x10a8bd0e0>]

As you can see, there's an extra body within the body. The resulting SOAP message looks like:

    DEBUG:zeep.transports:HTTP Post to https://opendap.co-ops.nos.noaa.gov/axis/services/ActiveStations:
    <?xml version='1.0' encoding='utf-8'?>
    <soap-env:Envelope xmlns:soap-env="http://schemas.xmlsoap.org/soap/envelope/">
      <soap-env:Body>
        <soap-env:Body/>
      </soap-env:Body>
    </soap-env:Envelope>

mvantellingen:
Should be fixed! Thanks

M1hacka:
Got the same error on 2.5.0 (current pypi version).
Sorry for no working example, I can't distribute it
Code:
```python
from zeep import Client
url = 'http://someurl'
client = Client(url)
auth_service = client.bind('SomeService')
# Expects xsd:string, xsd:string as AuthorisationRequest
auth_res = auth_service.AuthoriseUser('user', 'pwd')
```

Generated xml:
```xml
<?xml version='1.0' encoding='utf-8'?><br>
[...]

PR #480, opened 2017-06-23
Handle operations with no parameters properly.

dacut:
Possible fix for https://github.com/mvantellingen/python-zeep/issues/479


dacut:
I had to rework my initial patch -- a test caught that soap:Body is always required. I'm not happy with the revised patch -- there's a hack in there looking at self.body's tag in zeep.wsdl.messages.soap.SoapMessage.serialize(), preventing a soap:Body from rendering into itself.

mvantellingen:
Thanks, i actually missed this PR when looking at your issue. Solved it slightly different.

PR #495, opened 2017-07-10
return at least the type for empty elements

### LINKED ISSUE #494: Type is missing on empty elements
In wsdl:
```xml
<complexType name="SearchRealmsCommand">
    <complexContent>
        <extension base="tns2:TargetableCommand">
        <sequence>
        <element name="filter" nillable="true" type="tns14:Filter"/>
        <element name="limit" type="xsd:int"/>
        <element name="realms" nillable="true" type="impl:ArrayOf_tns149_RealmDTO"/>
        </sequence>
        </extension>
    </complexContent>
</complexType>
<complexType name="Filter">
    <sequence>
        <element name="comp" nillable="true" type="xsd:anyType"/>
    </sequence>
</complexType>
<complexType name="Empty">
    <sequence/>
</complexType>
<complexType name="Simple">
    <sequence>
        <element name="attr" nillable="true" type="xsd:string"/>
        <element name="core" type="xsd:boolean"/>
        <element name="type" nillable="true" type="xsd:anyType"/>
        <element name="value" nillable="true" type="xsd:anyType"/>
    </sequence>
</complexType>
<complexType name="Equal">
    <sequence/>
</complexType>
```

That request parameter "Filter" will be constructed like that:
```python
        search = self.client.type_factory("http://search.common.rsa.com")
        comp = search.Simple(core=True,
                             type=search.Equal(),
                             attr="name",
                             value=xsd.AnyObject(xsd.String(), "someNeedleString"))
        command = self.client.get_type("{http://admin.rsa.com}SearchRealmsCommand")(
            filter=search.Filter(comp=comp),
            limit=0,
        )
```
Backend expects (at least):
```xml
<filter>
    <comp xmlns:ns2="http://search.common.rsa.com" xsi:type="ns2:Simple">
    <attr>name</attr>
    <core>true</core>
    <type xsi:type="ns2:Equal"/>
    <value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:string">SystemDomain</value>
    </comp>
</filter>
<limit>0</limit>
<realms xsi:nil="true"/>
```
but without patch zeep is sending:
```xml
<filter>
    <comp xmlns:ns1="http://search.common.rsa.com" xsi:type="ns1:Simple">
    <attr>name</attr>
    <core>true</core>
    <type/>
    <value xmlns:xs="http://www.w3.org/2001/XMLSchema" xsi:type="xs:string">SystemDomain</value>
    </comp>
</filter>
<limit>0</limit>
<realms xsi:nil="true"/>
```
(which leads to fault java.lang.NullPointerException ooops ;) )

imho src/zeep/xsd/types/complex.py:render returns too early, so i moved the type setter upwards.

macrojames:
(Backend does casting/switch a filter logic based on the type)

fixes #494

### DIFF
--- a/src/zeep/xsd/types/complex.py
+++ b/src/zeep/xsd/types/complex.py
@@ -209,13 +209,12 @@ class ComplexType(AnyType):
         if not render_path:
             render_path = [self.name]
 
-        if xsd_type:
-            if xsd_type._xsd_name:
-                parent.set(xsi_ns('type'), xsd_type._xsd_name)
-            if xsd_type.qname:
-                parent.set(xsi_ns('type'), xsd_type.qname)
-
         if not self.elements_nested and not self.attributes:
+            if xsd_type:
+                if xsd_type._xsd_name:
+                    parent.set(xsi_ns('type'), xsd_type._xsd_name)
+                if xsd_type.qname:
+                    parent.set(xsi_ns('type'), xsd_type.qname)
             return
 
         if isinstance(value, ArrayValue):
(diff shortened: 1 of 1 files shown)


macrojames:
Please advise how to proceed.
The test are failing because some attribute order was changed and type is explicitly set now at least on complex elements. I would even add that to all elements but that might not be required.

I personally would change the tests.

PR #502, opened 2017-07-19
[FIX] soap envelope should be parsed with c14n

etobella:
When you are creating the signature, and exporting, it cannot validate, because it is formatting it. It should be encoded using c14n method.

### DIFF
--- a/src/zeep/wsdl/utils.py
+++ b/src/zeep/wsdl/utils.py
@@ -22,8 +22,7 @@ def get_or_create_header(envelope):
 
 
 def etree_to_string(node):
-    return etree.tostring(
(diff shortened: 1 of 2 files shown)


mvantellingen:
Thanks, I need to do some research for possible negative effects. Any hints?

etobella:
I am not sure which problems can it give to the system, but when I tried to send a Soap Message with a signature, it failed. However, I think that another options is to remove the  pretty_print=False option. When you are signing an xml, you cannot modify it, not even intros or spaces.

etobella:
I close this PR because I created a less agressive option #557

Issue #514, opened 2017-07-27
TransportError does not store error details inside


penguinolog:
TransportError exception stores only string ` 'Server returned HTTP status %d (%s)'`, while in several cases status code and content is useful for external applications.

Example case: Server return 401 error -> call login procedure and retry request.

mvantellingen:
This was merged

PR #515, opened 2017-07-27
Fix: TransportError does not store error details inside #514

penguinolog:
Add status_code and content optional fields. Add test for code 401.


mvantellingen:
Thanks!

PR #517, opened 2017-08-01
Fix: XMLSyntaxError does not store error details inside #514

penguinolog:
In some cases (load phase)
  XMLSyntaxError is raised instead of TransportError
  (Example URL:
   https://webservices2.autotask.net/ATServices/1.5/atws.asmx?WSDL
  )
Add content optional field.


mvantellingen:
Looks good, thanks!!

Issue #519, opened 2017-08-01
Async client do not raise HTTP errors from response


penguinolog:
Reproduce:
  Example URL: https://webservices2.autotask.net/ATServices/1.5/atws.asmx?WSDL
  Do not specify password
Actual status:
  HTTP Status 500
  XMLSyntaxError
Expected status:
  HTTP Status 500
  TransportError

penguinolog:
After migration to aiohttp 2.0+ will be possible to set `raise_for_status=True` session-wide.

PR #520, opened 2017-08-01
Fix: raise TransportError in aiohttp session #519

penguinolog:
Added test. (In test content is empty)


mvantellingen:
Thanks again !! Really appreciated! 👍

PR #530, opened 2017-08-08
Add Support For Tornado Async Transport

lakatalina8:
Pull request includes:
- support for tornado `gen.coroutine` async transport (non blocking I/O for Python 2.7+)
- light change in `wsdl.py` to allow transport customizations, it should be enough to inherit new transport class from `zeep.Transport` and link it with your bindings if necessary (didn't document that)
- unit tests for `TornadoAsyncTransport`
- setup and documentation updates

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -10,6 +10,7 @@ Highlights:
  * Support for Soap 1.1, Soap 1.2 and HTTP bindings
  * Support for WS-Addressing headers
  * Support for WSSE (UserNameToken / x.509 signing)
+ * Support for tornado async transport via gen.coroutine (Python 2.7+)
  * Support for asyncio via aiohttp (Python 3.5+)
  * Experimental support for XOP messages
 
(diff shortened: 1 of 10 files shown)


lakatalina8:
I am not sure that I understand how codecov detected -0.42% decrease in `src/zeep/transports.py`, as I have removed only the line there, so coverage should be potentially a bit higher...

mvantellingen:
This looks super neat! Thanks! I'm not really familiair with tornado so i'm trusting that this works 😛

lakatalina8:
You're welcome. Yes it works, I have tested gets/posts with live SOAP service and tornado, and usage of custom auth and client certs as well. But I'm converting `requests.Session` to `tornado.httpclient.HTTPRequest`, so it is possible that there could be some not supported scenarios. I think that it can be fixed/added on demand. You can add me to notifications list on Zeep issues creation, or reference me if issue is related to Tornado. Have a nice day ;-)

PR #555, opened 2017-09-02
Use __version__ from zeep instead of hardcoded string

gdalmau:
This is to change `version` parameter in `setup.py` file to use `__version__` from zeep dynamically

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@ import re
 import sys
 
 from setuptools import find_packages, setup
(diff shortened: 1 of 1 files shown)


justanr:
This'll break if any imports in init need anything from an uninstalled dependency.

mvantellingen:
This won't work indeed. Sorry closing for now as this is such a minor thing that it isn't worth it. The versions are automatically updated via versionbump

PR #557, opened 2017-09-04
[FIX] Delete pretty_print option

etobella:
I created a second option less agressive and it also works.


etobella:
After rechecking de code, I cannot reproduce again the problem with the original code. Maybe there was an issue with the certificates that I cannot understand yet :(.
I close the PR and if I find the problem again I will try to see what is really happening.

Issue #391, opened 2017-03-20
EntitiesForbidden


dr-leo:
Using zeep v1.3.0 checked out a few minutes ago, Win7x32, Python 3.5, the following script yields the above error:

import zeep
wcf = 'http://dataservices.imf.org/sdmx21/SDMX_WCF_Service21.svc'
url = 'http://dataservices.imf.org/sdmx21/SDMX_WCF_Service21.wsdl'
c = zeep.Client(wsdl=url)

dr-leo:
I consider this a bug, not a feature, because one wants to access data.imf.org.

Fixing it seems to be all about relaxing defusexml's security constraints. I have no experience with this. To minimize the trade-off, it is desirable to configure defusexml security at Client level or even method call level. But maybe there are other workarounds.

DrJackilD:
@dr-leo @mvantellingen I created PR for this feature, and would like to know your opinion about it. Thanks!

PR #567, opened 2017-09-15
Add ability to handle defusexml security rules when parse xml

DrJackilD:
Regard to issue [#391](https://github.com/mvantellingen/python-zeep/issues/391).

I really would like to use zeep in my project, because it has very friendly API and some other features, which is useful for me, but I can't do this, because zeep's Client don't have any tools to handle `defusexml` security rules, when using `fromstring` in the code. 

Generally, it's about behavior with `<!DOCUMENT>` and `<!ENTITIES>` tags in schema or xml. For example, I can't use `zeep` to work with `OpenTravel` wsdls, because they using DTD with entities inside wsdls. So, I've create PR, which add ability to handle this behaviour.

Will be happy to hear your things about it. Thanks!

### DIFF
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -115,12 +115,14 @@ class Client(object):
 
     def __init__(self, wsdl, wsse=None, transport=None,
                  service_name=None, port_name=None, plugins=None,
-                 strict=True, xml_huge_tree=False):
+                 strict=True, xml_huge_tree=False, forbid_dtd=False,
+                 forbid_entities=True):
         if not wsdl:
             raise ValueError("No URL given for the wsdl")
 
         self.transport = transport if transport is not None else Transport()
-        self.wsdl = Document(wsdl, self.transport, strict=strict)
+        self.wsdl = Document(wsdl, self.transport, strict=strict,
+                             forbid_dtd=forbid_dtd, forbid_entities=forbid_entities)
         self.wsse = wsse
         self.plugins = plugins if plugins is not None else []
         self.xml_huge_tree = xml_huge_tree
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -109,7 +109,10 @@ class Client(object):
     :param plugins: a list of Plugin instances
     :param xml_huge_tree: disable lxml/libxml2 security restrictions and
                           support very deep trees and very long text content
-
+    :param forbid_dtd: disallow XML with a <!DOCTYPE> processing instruction
+    :type forbid_dtd: bool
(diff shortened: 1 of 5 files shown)


sapphireua:
Cool stuff! I really appreciated with this! Would be great if it will be merged ASAP!

getser:
I've faced the issues described by DrJackilD too. Please accept this PR as soon as you can. Thanks in advance!

dr-leo:
Hi,

Thanks for this PR which seems to solve #391 I opened a few months ago.

However, I would make the following comments:

1. At first glance it appears you want to expose two flags from defusexml. But defusexml has at least one more such flag to configure its security level. Can I suggest to expose/propagate all config values defusexml has?

2. It seems you propose to relax the default behavior of defusexml. So you would relax the Default security level of zeep. I'd rather suggest not to Change zeep's default behavior in any way. Rather I'd prefer to simply allow the user to relax the security rules of defusexml/zeep.

If the above comments rest on incorrect assumptions, I apologize in advance.
 
(I'll Change the title of #391 to reflect the above.

Leo

DrJackilD:
@dr-leo Thanks for your comment! Ok, let me answer on your questions:
1. You're right, here is one more flag (`forbid_external`), I think I must add it here as well as other. Thanks for this note!
2. No, I don't propose it. This two flags are set to default as in `defusexml's` `fromstring`. Here is the reference from official documentation:
```
forbid_dtd (default: False)
disallow XML with a <!DOCTYPE> processing instruction and raise a DTDForbidden exception when a DTD processing instruction is found.

forbid_entities (default: True)
disallow XML with <!ENTITY> declarations inside the DTD and raise an EntitiesForbidden exception when an entity is declared.

forbid_external (default: True)
disallow any access to remote or local resources in external entities or DTD and raising an ExternalReferenceForbidden exception when a DTD or entity references an external resource.
```
So, as far as `zeep` using `fromstring` without any of this flags, the behavior will be the same, as previous, but with ability to relax security rules.

Best regards!

dr-leo:
Great!

I hope to give your Code a try within the next few days.

I still think there is merit in a way to get and set These flags for
an existing Client instance. This would require documented properties.
I haven't looked at your most recent Code, but I understand the only
API Extension you are proposing for now is in __init__. I would prefer
more Control and transparency. But we can defer this until Michael has
shared his General thoughts on this, I think, very useful and much
needed PR.

Leo

mvantellingen:
This should now be possible in master (will be 3.0):

```python
from zeep import Client, Settings

client = Client(wsdl, settings=Settings(forbid_entities=False)
```

Let me know if this isn't enough be opening a new issue

PR #569, opened 2017-09-18
corrected code typo and example

pashute:
Nested using value_1 example had 
a. maxOccurs="unbounded" (should be 1) 
b. type_2 not needed in example and makes for misunderstanding.

Nested list using value_1 example had typo - missing ]


mvantellingen:
Thakns

PR #572, opened 2017-09-21
Encode str in UnresolvedCustomeType in order to handle unicode names.

Thalmann:


Thalmann:
We need this fix in production / we are using it in production.. When do you have time to review?

Issue #257, opened 2016-11-15
Missing items when parsing Choice


ri8on:
- zeep 0.22
- [wsdl.zip](https://github.com/mvantellingen/python-zeep/files/591572/wsdl.zip)
- [script](https://gist.github.com/ri8on/38fcbcbd1edaac2977243b767f10d972)

response type looks like:
```
<xs:element name="getStateResult">
  <xs:complexType>
    <xs:complexContent>
      <xs:extension base="base:BaseAsyncResponseType">
        <xs:choice minOccurs="0">
          <xs:element ref="base:ErrorMessage"/>
          <xs:element name="NsiItem" type="nsi-base:NsiItemType"/>
          <xs:element name="NsiList" type="nsi-base:NsiListType"/>
        </xs:choice>
        <xs:attribute ref="base:version" use="required" fixed="10.0.1.2"/>
      </xs:extension>
    </xs:complexContent>
  </xs:complexType>
</xs:element>
```

`RequestState` and `MessageGUID` must be present, as defined in `BaseAsyncResponseType`, and one of `ErrorMessage`, `NsiItem` or `NsiList` can be present. However, only one (the first) element is in the output.

In the example the number of elements is limits here:
zeep.xsd.indicators.Choice#parse_xmlelements:
```
        if self.accepts_multiple:
            result = {name: result}
        else:
            result = result[0] if result else {}
```

mvantellingen:
Hi @ri8on, could you give the master branch a try? It should be fixed now. Thanks btw for the all the provided data! :-)

mvantellingen:
I'm closing this issue for now. If in new releases this is still a problem then please let me know so that I can re-open it.

Thanks again for taking the time to report the issue

766dt:
Thanks, the issue now is not reproduced anymore.
I apologize for the delay, and thank you for your work! =)

ri8on:
Thanks, the issue now is not reproduced anymore.
I apologize for the delay, and thank you for your work! =)

PR #597, opened 2017-10-26
fix extension of base type containing choice by the type containing sequence.

toshka:
This is similar to #257 but from the other side.
If you try to extend type with choice element by the type with sequence all items from choice would be copied to sequence and became flatten.
Look into the schema from test_choice_extend_base test in all became clear.


mvantellingen:
Merged, thanks!!

PR #604, opened 2017-11-03
Deal with Nil on a complexType with simpleContent

flupzor:
When Nil is set on an element which is built up from a complexType with simpleContent don't render it as a element with the text content 'Nil', but as the xsi:nil attribute on the element.


mvantellingen:
thanks!

PR #627, opened 2017-11-30
Added proxy feature

f0cker:
Ability to use proxy (http/s/socks5) for sending all requests. Also allowed for a custom user-agent to be used.

Example usage:
client = Client(wsdl=wsdl, strict=False, transport=Transport(proxy_prot='http', proxy_server='http://127.0.0.1:8080', user_agent='Firefox/5.0'))


mvantellingen:
I think this can be done outside of zeep by creating a request session object with the correct proxy settings and pass that to zeep. See for example http://docs.python-zeep.org/en/master/transport.html#ssl-verification

f0cker:
You're right, thanks for pointing that out, it works fine. Could you add the lines to make the User-Agent optional?

PR #629, opened 2017-12-03
Made maxOccurs validation exception say the correct expected number. …

wesleylima:
…Showing the invalid number found in response in exception.


mvantellingen:
👍

PR #638, opened 2017-12-10
Added try, exept to deserialization of body.

jring:
When server returns simple value (non array) calling len() will cast an exception.


mvantellingen:
Thanks for your PR. Are you able to add a unittest for this change? This allows me to easily confirm which bug is fixed and that it isn't prone to a regression in the future.

PR #646, opened 2017-12-21
Fix Error when a name contains a space

cesarlvielma:
I found an API that contains spaces in their fields names like:

<>NAME <>

and generates this error when the XML was parsing:
File "src/lxml/apihelpers.pxi", line 1631, in lxml.etree._tagValidOrRaise (src/lxml/etree.c:35382)
ValueError: Invalid tag name u'NAME '

This fix strip all spaces in the field names.


mvantellingen:
I think this is already fixed in master

PR #661, opened 2018-01-16
Add connect_timeout to TornadoAsyncTransport requests to match sync Transport behaviour

libcthorne:
Zeep's synchronous Transport class uses Requests' timeout parameter to support `operation_timeout` and `load_timeout`:

- https://github.com/mvantellingen/python-zeep/blob/master/src/zeep/transports.py#L42
- https://github.com/mvantellingen/python-zeep/blob/master/src/zeep/transports.py#L125

Passing in a number as the timeout will result in the connect timeout and request timeout being set to this number. ([docs](http://docs.python-requests.org/en/master/user/advanced/#timeouts), [code](https://github.com/requests/requests/blob/d55de2d391d1350766221ce0aa36e670dc2a0169/requests/adapters.py#L418))

> If you specify a single value for the timeout, like this:
> r = requests.get('https://github.com', timeout=5)
> The timeout value will be applied to both the connect and the read timeouts.

However, Zeep's Tornado Transport class uses Tornado's request_timeout parameter:

- https://github.com/mvantellingen/python-zeep/blob/master/src/zeep/tornado/transport.py#L41
- https://github.com/mvantellingen/python-zeep/blob/master/src/zeep/tornado/transport.py#L109

This doesn't set a connection timeout and results in Tornado's [default connect_timeout value of 20 seconds](http://www.tornadoweb.org/en/stable/httpclient.html#tornado.httpclient.HTTPRequest) being used.

The effect of this is that `operation_timeout` or `load_timeout` values greater than 20 seconds used with Tornado's transport can appear to be ignored if a connection takes a long time. This PR proposes to set connect_timeout in order to be consistent with the synchronous transport's behaviour.

### DIFF
--- a/src/zeep/tornado/transport.py
+++ b/src/zeep/tornado/transport.py
@@ -38,6 +38,7 @@ class TornadoAsyncTransport(Transport):
         client = httpclient.HTTPClient()
         kwargs = {
             'method': 'GET',
+            'connect_timeout': self.load_timeout,
             'request_timeout': self.load_timeout
         }
         http_req = httpclient.HTTPRequest(url, **kwargs)
@@ -106,6 +107,7 @@ class TornadoAsyncTransport(Transport):
 
         kwargs = {
             'method': method,
+            'connect_timeout': self.operation_timeout,
             'request_timeout': self.operation_timeout,
             'headers': dict(headers, **session_headers),
             'auth_username': auth_username,


mvantellingen:
Thanks! I haven't done a lot with tornado so i'm trusting you on this one :-)

PR #667, opened 2018-01-23
Added example for trafficvance using apikey

MichaelSvendsen:
After some challenges creating a working solution for applying apikey to trafficvance we got this working.


mvantellingen:
Great, thanks

PR #681, opened 2018-02-08
Clarify certificate usage in transport docs

guyfawcus:
Add note about requests.Session.cert

Not having worked with certs much in the past, the current transport docs lead me a bit astray. I kept trying to use `verify` whereas  I should have been using `cert`.

I'm still not sure if this is the best way to get the message across but I reckon it's more helpful than before.
(Please feel free to rip this apart, just wanted to spread the word of a stumbling block for newcomers)


mvantellingen:
Thanks! It's a bit tricky to document just enough without duplicating the complete requests doc. Perhaps I need to a add more references to the docs since the transport is just a simple wrapper around requests

PR #690, opened 2018-02-16
Add check to handle simple return values.

kompi:
Previously, returned values that didn't have a 'len' method raised TypeError:
Example:
The corresponding WSDL section:
<xsd:element name="DisconnectResponse" nillable="false" type="tns:int"/>
After calling 'disconnect':
TypeError: object of type 'int' has no len()

### DIFF
--- a/CONTRIBUTORS.rst
+++ b/CONTRIBUTORS.rst
@@ -50,5 +50,4 @@ Contributors
 * Raymond Piller
 * Zoltan Benedek
 * Øyvind Heddeland Instefjord
-
-
+* Adam J. Varga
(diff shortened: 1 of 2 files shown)


kompi:
> Also running in this bug. Is there something I can do to help this along?

No idea mate, I created the PR to fix it, but it was ignored.

ron819:
@mvantellingen can you check this one?

lykinsbd:
Please do not merge this PR, it will break if any response is None, removing current functionality.

    >>> blah = None
    >>> len(blah)
    Traceback (most recent call last):
      File "<input>", line 1, in <module>

I will look at submitting a modified version of this PR that keeps the `if result is None` portion of the check too.

PR #695, opened 2018-02-27
Use __getattr__() for CompoundValue instead of __getattribute__()

vdboor:
The latter is a bit less efficient, and triggered for every attribute, while __getattr__() is only called for missing attributes - which is exactly what is needed to read the SOAP object attributes as normal object attributes.

The build errors are due to a version conflict with aiohttp, and not caused by this patch.


mvantellingen:
The idea was to only use ``__values__`` as location for attributes. This changes that i belive

vdboor:
I doubt so. `__getattr__()` is triggered for every missing value, and doesn't change the object nor store values. When `__getattribute__()` is overwritten, every attribute access is proxied through, even for existing values.

PR #713, opened 2018-03-22
Fix compatability with aiohttp>=3

atx:
Note that this might be a breaking change - not sure what your policy is on that.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ tests_require = [
 
 if sys.version_info > (3, 4, 2):
     async_require.append('aiohttp>=1.0')
-    tests_require.append('aioresponses>=0.1.3')
+    tests_require.append('aioresponses>=0.4.1')
 
 
 with open('README.rst') as fh:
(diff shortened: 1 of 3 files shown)


decaz:
@atx can you make some fixes so tests will pass?

atx:
Oops, completely forgot about this - fixed.

The remaning test failures seem to be [caused](https://github.com/jupyter/notebook/issues/3397) by Tornado 5.0 (testing with 4.5.3 works).

vdboor:
@atx is there any advantage of specifying the timeout over what I've done in PR #735?

atx:
I don't think there is. aiohttp itself does some [more complicated stuff](https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L181) with its timeout code but it's not obvious to me if it really matters.

mvantellingen:
I think this was a duplicate with #735 right? I've merged that one before reviewing this one (sorry)

dirkmb:
besides the fact that #735 missed to change the `get`-method of `AsyncTransport`, which still uses `aiohttp.Timeout`

Issue #658, opened 2018-01-09
Weird behaviour of some calls


acidjunk:
We replaced SUDS with Zeep and are very happy with the results, kudo's for this excellent lib!

The SOAP server facilitates some complexer call's to search in data: these call's seem to violate some rules in Soap (or in Zeep) so depending on the endpoint (table); we get results or not. I needed to disable strict mode to get results at all.

The problem seems related to the return types that are links to the original content. I also included an response that shows how a normal response with multiple items is facilitated. Zeep fails on some of the responses for the find. I noticed 2 cases:
a) Data isn't available in an soap object attribute att all but in an attribute: _raw_elements. (wrong_request.xml and wrong_response.xml)
b) Data is available in the normal way. (correct_request.xml and correct_response.xml)

For both cases the SUDS based solution returns SOAP objects that mimic the behaviour of a call to GetProductById or GetAllProducts: e.g. it would return Products instead of links to Products.

Mandatory information:
1. The version of: tested with zeep 2.4.0 & 2.5.0
2. The WSDL you are using: see https://github.com/acidjunk/zeep_example/tree/master/wsdl
3. An example script: see https://github.com/acidjunk/zeep_example/blob/master/an_example_script.py

acidjunk:
When I enable strict mode:
Unexpected element '{http://schemas.datacontract.org/2004/07/PresentationModel}Errors'

But that seems just the first error. Is there anything I can do to help debugging the behaviour?

acidjunk:
It seems related to how stuff is defined in the XSD's. But it looks like zeep is do-ing some name space stuff that doesn't handle this case.

First a GOOD example: Port
```
    <xs:complexType name="Port">
        <xs:complexContent mixed="false">
            <xs:extension base="q1:BaseObject" xmlns:q1="http://schemas.datacontract.org/2004/07/PresentationModel">
                <xs:sequence>
                    <xs:element minOccurs="0" name="Blocked" nillable="true" type="xs:string"/>
                    <xs:element minOccurs="0" name="CableCircuit" nillable="true" type="q2:Circuit"
                                xmlns:q2="http://schemas.datacontract.org/2004/07/PresentationModel.Circuits"/>
                    <xs:element minOccurs="0" name="CableCircuitId" nillable="true" type="xs:int"/>
                    <xs:element minOccurs="0" name="CableCircuitPathDetails" nillable="true"
                                type="q3:CableCircuitPathDetails"
[...]

PR #714, opened 2018-03-24
A pull request that shows one failing test and one passing test for issue: #658

acidjunk:
As requested in #658 : a failing test.


acidjunk:
@mvantellingen Now the same test fail that are also already failing in master. The tests contain one test that succeeds test_find_customer() and one test that fails test_find_product().

mvantellingen:
Hi, thanks for this. Is it possible to reduce the test code so we don't need to add multiple xsd files?

acidjunk:
Yes that should be possible: but I wonder if that 's not the root cause of this issue:
IMHO the bug itself was introduced in a commit that tries to solve another (regression)bug.

I can confirm that the current behaviour is an regression though: version 0.27, 1.0, 1.1 and 1.2 work ok (strict mode enabled from v 1.1)

from version 1.3 and up it returns parse errors. Luckily this narrows it down a lot as there is only 2 days between version 1.2 and 1.3 :-)

AFAIK the regression is introduced in commit: 349c27c
The last OK commit: bdb2a14

As I don't know enough about XSD/nesting/complex types I'm having an hard time to actual understand what problem/regression 1.2 had that was supposedly fixed in 1.3? Can I trust the facked XSD? And the way I see it, the bug will be somewhere in XSD's that inherit more then once from a complex type. Not for all of these constructs: as in my example some of the call's work ok: so I assume it also is related to the order of XSD object definitions.

For now we use my fork with a fix: I'm willing to improve the test in my free time (or to add a new one) when I get a bit clarity of:
- the actual regression that 1.3 fixes
- chances that this fix will result into a release (so I can remove the fork from our requirements)

hmvp:
@mvantellingen I rebased the original fix and updated the testcase to be a lot smaller. However the original fix (as found in https://github.com/acidjunk/python-zeep/blob/original_fix/src/zeep/xsd/types/complex.py) does not work anymore in 4.0.0

Would you be so kind to check why the second assert of the tests fail and see if you can fix that?

hmvp:
> @mvantellingen I rebased the original fix and updated the testcase to be a lot smaller. However the original fix (as found in https://github.com/acidjunk/python-zeep/blob/original_fix/src/zeep/xsd/types/complex.py) does not work anymore in 4.0.0
> 
> Would you be so kind to check why the second assert of the tests fail and see if you can fix that?

Hmm I might have simplfied to much due to `xsd_ignore_sequence_order=False` let me check my testcase again

hmvp:
Had to slightly tweak the test (I'm sorry for the huge wsdl, but reducing somehow removed the error)

Also added a fix for some existing issues in the master branch feel free to ignore if not wanted

acidjunk:
Let me know if there is anything that could be done to speed things up. The bug and a wrong test was introduced after a very short described issue on which the author of the issue never replied); so the process to get this accepted feels a bit ironic. If there is a way to sponsor/donate that has a positive effect on the speed : that would be very nice.

PR #719, opened 2018-04-02
Space issues with zeep for xml tags fixed

galacticsurfer:


mvantellingen:
Manually applied since your PR had some other changes, also added unittest
Thanks!

PR #723, opened 2018-04-03
Fix proposal for #722

### LINKED ISSUE #722: Includes with no default namespace in parent schema not resolving nested types
When there's no default namespace in parent schema (the one with include tag), then nested types in included schema fail to resolve. 

[This unit test](https://github.com/kuba-baku/python-zeep/blob/39949da90730ebd880095103f128337f01fa0bef/tests/test_xsd_schemas.py#L729) fails with:
`zeep.exceptions.NamespaceError: Unable to resolve type itemType. No schema available for the namespace None.`

The schema is valid, verified using lxml.

It's recommended to set default NS in imported schema to targetNS from parent schema, see this [W3C reccomendation for Chameleon Inclusion](https://www.w3.org/TR/xmlschema11-1/#chameleon-xslt)

Forked from e4a12d81d573dafbe1f0c38671835d742fa5fc83

kuba-baku:
Trying to fix #722 by setting default NS in imported schema according to targetNS from parent schema.
Tested only on Python 3.5.

### DIFF
--- a/src/zeep/xsd/visitor.py
+++ b/src/zeep/xsd/visitor.py
@@ -247,14 +247,20 @@ class SchemaVisitor(object):
         # transfer the default namespace to the included schema. We can't
         # update the nsmap of elements in lxml so we create a new schema with
         # the correct nsmap and move all the content there.
-        if not schema_node.nsmap.get(None) and node.nsmap.get(None):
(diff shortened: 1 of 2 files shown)


mvantellingen:
Merged, thanks!

PR #748, opened 2018-05-08
Process errors prior to verifying wsse

vincecarney:
Attempting an invalid Login() action on a service will not have a valid security header. I moved the error processing above the wsse verification to properly handle the fault codes.


mvantellingen:
Shouldn't we verify the wsse on fault responses too?

PR #750, opened 2018-05-12
Allow headers to optionally be returned from client.create_message.

havocbane:
I noticed that the http headers were already being returned from the binding._create call, but were being discarded. I have use for these, so I would like to be able to have them conditionally returned with a new kwarg.

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -32,6 +32,26 @@ def test_bind_service_port():
     assert service
 
 
+def test_create_message():
+    client_obj = client.Client('tests/wsdl_files/soap.wsdl')
+    service = client_obj.bind('StockQuoteService')
+    assert service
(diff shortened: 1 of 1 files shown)


src/zeep/client.py
```diff
@@ -121,15 +121,19 @@ def create_service(self, binding_name, address):
                 "are: %s" % (', '.join(self.wsdl.bindings.keys())))
         return ServiceProxy(self, binding, address=address)
 
-    def create_message(self, service, operation_name, *args, **kwargs):
+    def create_message(self, service, operation_name, *args, return_headers=False, **kwargs):
```
lucaspolo: Hello @havocbane, you must put the **return_headers=False** before ***args**

havocbane:
I'll add more test coverage over the changes.

PR #753, opened 2018-05-17
Added check for nil in ComplexType

mcwladkoe:
parse_xmlelement  "<element xsi: nil = "true" />" returns empty object not None


mcwladkoe:
It seems to me that there was a bug, because with the nullable instance in WSDL the wrong result is returned (with allow_nullable = True, it is expected to None, but we get an empty instance)
The solution can be such, or it is necessary to call the parent function from AnyType.

mvantellingen:
Can you add a unittest?

mcwladkoe:
Added test_xml_simple_content_nil_allow_nil
if written logic contradicts yours - I wait for suggestions

artur-martirosyan:
I had the same issue and it is also described in https://github.com/mvantellingen/python-zeep/issues/823
Was going to submit a fix then realized it would duplicate this one.
@mcwladkoe can you please merge from master and resubmit your PR?
Thanks

mcwladkoe:
@mvantellingen I've updated to the latest master, could you please review this Pull Request once again?

g1sky:
@mvantellingen The problem still occurs. After copying the code from this pull request into zeep library, the xml parses correctly. Could you please merge it? I need this change in my project.

PR #761, opened 2018-05-25
Fixed handling of minOrrurs of elements with @ref inside of choice

spilikin:
Hi,

when handling the XSD with referenced elements inside of choice the min_occurs was set to 1 and led to parsing errors. Since the code 
'''
            if parent.tag == tags.choice:
                min_occurs = 0
'''

was executed AFTER the element reference was resolved, the parent tag was always "schema" and never "choice". I moved the mic_occurs to the place where it is executed before the reference is resolved.

I hope the FIX is right and you can accept the Pull Request.

Cheers

### DIFF
--- a/src/zeep/xsd/visitor.py
+++ b/src/zeep/xsd/visitor.py
@@ -319,6 +319,10 @@ class SchemaVisitor(object):
         # be present. Short circuit that here.
         # Ref is prohibited on global elements (parent = schema)
         if not is_global:
+            # Naive workaround to mark fields which are part of a choice element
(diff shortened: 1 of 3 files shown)


mvantellingen:
Cool, Thanks. Are you able to add  a unittest?

mvantellingen:
Cool, Thanks. Are you able to add  a unittest?

spilikin:
Sure, will make one and pull again

spilikin:
unit tests are in.

PR #787, opened 2018-07-19
using strict from settings to parse xml

freakypie:
I have SOAP response that generally isn't valid XML, but I can't change it. Please lemme know if you know a better way than using `strict`


mvantellingen:
Just interested, is this also an AXIS 1.4 server?

mvantellingen:
I fixed it slightly different, see f49d78806cf1c5e5ce6a62c84a1fe375fd74618b

Thanks for the PR !

freakypie:
Thanks for zeep!

Issue #788, opened 2018-07-20
Error when trying UsernameToken and Signature together and signing. 


mitchea5:
Please provide the following information:

1. The version of zeep (or if you are running master the commit hash/date)
    3.0.0
2. The WSDL you are using
`   https://community.workday.com/sites/default/files/file-hosting/productionapi/Integrations/v29.1/Integrations.wsdl`

I am attempting to connect to an endpoint that requires a username and certificate authentication. 

There are two issues I am running into so far:

1) For some reason the userid was not being added so the previous developer was trying to add it via  plugin:

 ```
 session = Session()
    session.verify = 'pubkey'
    transport = Transport(session=session)
    client = Client(
        wsdl=myurl, transport=transport,
        wsse=Signature('privkey,'pubkey),
        plugins=[RequestHeaderPlugin()],
        )

```
The plugin adds the username token (and password if in debug mode).

```
 class RequestHeaderPlugin(Plugin):
        """
        Modify the request header and add Username and Password to the Security header created by
        the wsse:Signature call
        """
        def egress(self, envelope, http_headers, operation, binding_options):
            maker = ElementMaker(
                namespace='http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd')
            if args.nopasswd:
                node = maker.UsernameToken(
                    maker.Username(userid),
                )
            else:
                node = maker.UsernameToken(
                    maker.Username(userid),
                    maker.Password(password) ,
                )
            security = get_security_header(envelope)
            security.append(node)

```

When I specify the no password option I get the following message back:

`<SOAP-ENV:Envelope xmlns:SOAP-ENV="http://schemas.xmlsoap.org/soap/envelope/"><SOAP-ENV:Body><SOAP-ENV:Fault xmlns:wd="urn:com.workday/bsvc"><faultcode>SOAP-ENV:Client.authenticationError</faultcode><faultstring>invalid username or password</faultstring></SOAP-ENV:Fault></SOAP-ENV:Body></SOAP-ENV:Envelope>`

So reviewing this, it sounds like it is not necessary to do this, this should work:

`http://docs.python-zeep.org/en/master/wsse.html?highlight=x509#usernametoken-and-signature-together`

So:

 ```
  session = Session()
    session.verify =pubkey
    user_name_token = UsernameToken(userid)
    signature = Signature(privkey,pubkey),
    transport = Transport(session=session)
    client = Client(
[...]

PR #789, opened 2018-07-21
Support xmlsec cred + signature pairing in verification stage

tonybaloney:
Relates to #788

### DIFF
--- a/src/zeep/wsdl/bindings/soap.py
+++ b/src/zeep/wsdl/bindings/soap.py
@@ -181,7 +181,11 @@ class SoapBinding(Binding):
                 message_pack = None
 
         if client.wsse:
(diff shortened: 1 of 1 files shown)


mvantellingen:
Nice one, thanks. To make sure that this keeps working in future releases, could you please add a unittest for this? (as a separate PR?)

tonybaloney:
@mvantellingen sure can!
it doesn't look like this feature has any tests so I'd be creating a new test module for sig+creds.
`test_wsse_username` has a `test_integration` method which instantiates a client object and mocks out the HTTP response, I'd need to do something similar with signatures.
However, `test_wsse_signature` doesn't have an integration test like this.

Any advice on which tests I should put in place? I was going to first add a test that demonstrates the error

PR #792, opened 2018-07-27
Fix SSL issue on with `TornadoAsyncTransport`

shlomi-shalem:
I tried to use the `tornado` module with HTTPS, and got the following error message:

```
TypeError: coercing to Unicode: need string or buffer, bool found
```

After digging the code a little bit, I found that `ca_certs` is not being passed right to the `HTTPRequest` object inside `zeep.tornado.TornadoAsyncTransport.fetch`. From the documentation of `HTTPRequest`:

```
:arg str ca_certs: filename of CA certificates in PEM format,
           or None to use defaults.
```

Yet, the code provide a `bool` value to it:

```python
            'ca_certs': self.session.verify,
```

Fixed this (and another small bug with `validate_cert`) and it seems to work right for me now.

### DIFF
--- a/src/zeep/tornado/transport.py
+++ b/src/zeep/tornado/transport.py
@@ -95,10 +95,11 @@ class TornadoAsyncTransport(Transport):
         # extracting client cert
         client_cert = None
         client_key = None
+        ca_certs = None
 
         if self.session.cert:
             if type(self.session.cert) is str:
-                client_cert = self.session.cert
+                ca_certs = self.session.cert
             elif type(self.session.cert) is tuple:
                 client_cert = self.session.cert[0]
                 client_key = self.session.cert[1]
@@ -113,8 +114,8 @@ class TornadoAsyncTransport(Transport):
             'auth_username': auth_username,
             'auth_password': auth_password,
             'auth_mode': auth_mode,
-            'validate_cert': self.session.verify is not None,
-            'ca_certs': self.session.verify,
+            'validate_cert': bool(self.session.verify),
+            'ca_certs': ca_certs,
             'client_key': client_key,
             'client_cert': client_cert
         }


shlomi-shalem:
The PR is actually ready for a review, but there seems to be a problem with the CI script:

```bash
3.7 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.7.tar.bz2
$ curl -sSf -o python-3.7.tar.bz2 ${archive_url}
curl: (22) The requested URL returned error: 403 Forbidden
Unable to download 3.7 archive. The archive may not exist. Please consider a different version.
```

I hope that it'll be fixed soon (I have ran the tests on Python 3.7 locally, and it went OK).

src/zeep/tornado/transport.py
```diff
@@ -113,8 +114,8 @@ def fetch(self, address, method, headers, message=None):
             'auth_username': auth_username,
             'auth_password': auth_password,
             'auth_mode': auth_mode,
-            'validate_cert': self.session.verify is not None,
-            'ca_certs': self.session.verify,
+            'validate_cert': bool(self.session.verify),
```
shlomi-shalem: The problem with the removed line is:
```python
False is not None
```
 will return `True`, which is not the wanted result.

shlomi-shalem:
Ready for review the Travis problem will go away. Thanks!

src/zeep/tornado/transport.py
```diff
@@ -113,8 +114,8 @@ def fetch(self, address, method, headers, message=None):
             'auth_username': auth_username,
             'auth_password': auth_password,
             'auth_mode': auth_mode,
-            'validate_cert': self.session.verify is not None,
```
shlomi-shalem:
```python
False is not None
```
 will return `True`, which is not the wanted result.

mvantellingen:
I've never used the tornado transport myself, but it looks good to me. So i'm going ahead wit the merge :-)

Thanks!

PR #795, opened 2018-07-30
`TornadoAsyncTransport` bug fix

shlomi-shalem:
Unlike `requests` package, Tornado's `AsyncHTTPClient` default behavior is to raise an exception in case of HTTP error. Therefore, in case of any error returned from the server, an `HTTPError` object is raised prior to `zeep.exceptions.Fault`. Fixed by suppressing this default behavior.


shlomi-shalem:
Fixed and tested. Ready to merge if interested. Thanks!

mvantellingen:
Thanks. There is no unittest added for this so it might break again in the future.

PR #799, opened 2018-08-07
Add an option to send a nonce on plaintext auth

nitbix:
In some cases (e.g. Oracle Primavera) a server expects a UsernameToken
in plain text, but containing a nonce. This adds an option to UsernameToken
(plaintext_nonce) which will send a nonce and a creation timestamp.


kaiserxc:
I would definitely second this. Also thanks for pointing the issue.

kaiserxc:
@nitbix I'm having issues using your fork to authenticate a P6 web service. Would you mind providing the code you used to access your P6 WS?

Thanks in advance.

PR #808, opened 2018-08-16
Fill timestamp

tpazderka:
Fill timestamp with `Created` and `Expires` elements.

### DIFF
--- a/src/zeep/wsse/signature.py
+++ b/src/zeep/wsse/signature.py
@@ -8,6 +8,7 @@ admittedly painful, will likely assist in understanding the code in this
 module.
 
 """
+import datetime
 from lxml import etree
 from lxml.etree import QName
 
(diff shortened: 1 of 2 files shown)


mvantellingen:
Could you add a unittest? And why exactly 5 minutes as ttl?

tpazderka:
Sure, I will add unittests. 5 minutes was my first ide, we can probable set the default ot 1 hour and leave it configurable through `Settings` object?

mvantellingen:
I think the expires value should be a setting. A default of 5 minutes or 60 minutes is fine by me

mvantellingen:
Wondering how this compares to #817

tpazderka:
In #817 the Timestamp token has to be created explicitely outside the call. This tales care of that in the code.

mvantellingen:
So I just merged #950 since that seemed sane, but then looked at this again.
Perhaps a proper solution is adding a kwarg to `Signature` like `Signature(timestamp=True)` ?

tpazderka:
I agree that #950 looks sane. I do believe that the timestamp should be signed if present so probably no changes needed?

PR #812, opened 2018-08-20
Implement BinarySignature

tpazderka:
This method works similarly to Signature but places the key as a
BinarySecurityToken


mvantellingen:
Thanks!

PR #817, opened 2018-08-28
Update username.py

vishu-guntupalli:
I use zeep on a daily basis to interact with some Oracle cloud SaaS web-services. The ws-security policy for Oracle cloud is to send a mandatory timestamp token with time to live. 

Adding timestamp token to WSS header.


mvantellingen:
Could you add a unittest?

vishu-guntupalli:
I added a test case, please verify.

PR #822, opened 2018-08-31
 Add support for different signing and verification certificates

tomerazran:
I added support for situations where the signing certificate is different from the response signature verification certificate.
The code added will is fully compatible with older versions, since the signature verification certificate will only be used if it is passed to the constructor - otherwise the signing certificate is used (same as used to be before the change).

### DIFF
--- a/src/zeep/wsse/signature.py
+++ b/src/zeep/wsse/signature.py
@@ -51,7 +51,7 @@ class MemorySignature(object):
         self.signing_key_data = signing_key_data
         self.signing_cert_data = signing_cert_data
         self.password = password
-        self.signature_verification_cert_data = signature_verification_cert_data
+        self.signature_verification_cert_data = signature_verification_cert_data if signature_verification_cert_data is not None else signing_cert_data
 
     def apply(self, envelope, headers):
         key = _make_sign_key(self.signing_key_data, self.signing_cert_data, self.password)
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks. I don't think it's fully backwards compatible because kwargs are renamed. If we can keep those the same as now and only add a new optional keyword argument then i can merge it

zanchey:
Will there ever be a case, other than testing via round-trip, where the certificate used for sending messages is the same as the one that signs received messages?

mvantellingen:
@zanchey I build this originally for a use case were it was the case that the certificates were the same. If that is best practice I'm not sure :-)

Either way, it should be pretty easy to keep this backwards compatible by not renaming the kwargs

gil-obradors:
Hi!

I confirm that this scenario is a real with a National Telecom Webservice.
Would be nice to pull.
If anyone can mentor me, I will do the work as the documentation.

Thanks to all

epetrovski:
I have the same issue with a Government SOAP api - signing and verification certificates are different. Any plans for this?

epetrovski:
Using the `BinarySignatureTimestamp()` class from https://github.com/mvantellingen/python-zeep/issues/996#issuecomment-524173570 fixed the issue for me.

PR #836, opened 2018-10-02
Fix Transport.post log message response content decoding

lamenezes:
When log level is set to DEBUG and the webservice returns its response in a content-type different from utf-8, the following exception happens:

```
Traceback (most recent call last):
  ...
/lib/python3.7/site-packages/zeep/transports.py", line 78, in post
    log_message = log_message.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 1432: invalid continuation byte
```

(some of the traceback was deleted for readability purposes)

This problem happens because the response enconding is ignored when decoding the response content from bytes to str. In the example above the response encoding is `iso-8859-1`. And this PR solves this issue.

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -75,7 +75,7 @@ class Transport(object):
             else:
                 log_message = response.content
                 if isinstance(log_message, bytes):
-                    log_message = log_message.decode('utf-8')
+                    log_message = log_message.decode(response.encoding or 'utf-8')
 
             self.logger.debug(
                 "HTTP Response from %s (status: %d):\n%s",


mvantellingen:
👍

alfremedpal:
Hello! Kind of late to the thread but I didn't want to start a new one since this seems to be the origin of my problem. So, is there a way to turn off the internal logger? I'm using zeep 3.4.0 and I'm still getting the `UnicodeDecodeError` from this same line this PR fixed.

```
self.client.service.consultaXML(payload)\n",
        "  File \"/var/task/zeep/proxy.py\", line 45, in __call__\n    kwargs,\n",
        "  File \"/var/task/zeep/wsdl/bindings/soap.py\", line 122, in send\n    response = client.transport.post_xml(options[\"address\"], envelope, http_headers)\n",
        "  File \"/var/task/zeep/transports.py\", line 95, in post_xml\n    return self.post(address, message, headers)\n",
        "  File \"/var/task/zeep/transports.py\", line 75, in post\n    log_message = log_message.decode(response.encoding or \"utf-8\")\n",
        "UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd3 in position 1339: invalid continuation byte\n"
```

Which I guess could be turned off if I set my logger to something above `DEBUG`, however, if I set it to something higher I get a `TransportError`

```
TransportError: Server returned response (200) with invalid XML: Invalid XML content received (Input is not proper UTF-8, indicate encoding !
```

And it only happens with some responses, not every one of them. So, I'm confused, is the logger raising a decode error (when I set the logging level to `DEBUG`) or is the XML invalid (when there's no logger)?

@lamenezes @mvantellingen

PR #841, opened 2018-10-04
Xmlsec requirement

plaes:
Improve "developer experience" when xmlsec library is not installed.

### DIFF
--- a/tests/test_wsse_signature.py
+++ b/tests/test_wsse_signature.py
@@ -16,10 +16,14 @@ KEY_FILE = os.path.join(
 KEY_FILE_PW = os.path.join(
     os.path.dirname(os.path.realpath(__file__)), 'cert_valid_pw.pem')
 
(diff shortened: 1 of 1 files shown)


tests/test_wsse_signature.py
```diff
@@ -16,10 +16,14 @@
...
 
-skip_on_win32 = pytest.mark.skipif(sys.platform == 'win32',
-                                   reason="does not run on windows")
-
-@skip_on_win32
+# Check whether xmlsec library has been installed
+from zeep.wsse.signature import xmlsec as xmlsec_installed
+skip_if_no_xmlsec = pytest.mark.skipif(sys.platform == 'win32',
+                                       reason="does not run on windows") and \
+                    pytest.mark.skipif(xmlsec_installed is None,
```
plaes: It's also possible to use `pytest.importorskip('xmlsec')` ;)

PR #844, opened 2018-10-09
Bunch of improvements to Transport docs

plaes:
Bunch of improvements to Transport docs:
* Fix hint formatting
* Show easier way to disable tls/ssl verification
* Mention that SSL is deprecated and use TSL instead
* Add section about proxy usage


mvantellingen:
Thanks!

PR #846, opened 2018-10-10
docs: client: Add a section on how to load existing responses into XML

plaes:
I ran into this issue and needed to dig into code.. :)


mvantellingen:
I'm not sure we should be documenting private api. If this is functionality we should have then it should have a proper pubic api

PR #863, opened 2018-11-16
Add verify_ssl and proxy to AsyncTransport

azinoviev:


mvantellingen:
Merged, thanks.

Note that this might break again later since there are no unittests. So if you will use this in production then unittests are handy to add in a separate PR :-)

PR #873, opened 2018-11-21
Fix bad reference in BinarySignature

tpazderka:


mvantellingen:
Can you explain what this fixes?

tpazderka:
If I recall correctly, the `BinarySecurityToken` was not referenced. The reference was the other way around.

mvantellingen:
ok, I think it makes sense, and it doesn't seem to break the tests

Issue #499, opened 2017-07-17
Improve handling of out-of-order values in a sequence


nmaludy:
### The version of zeep (or if you are running master the commit hash/date)
```
$ pip show zeep
Name: zeep
Version: 2.2.0
```

### The WSDL you are using
https://gist.github.com/nmaludy/5b44b6b8ce6e0151481250521283d684

### An example script
https://gist.github.com/nmaludy/e467d0e0d223d3fe0a9e49a424a08826

### XML returned from GetRanges() call
https://gist.github.com/nmaludy/0a22fd84d4c41a7ad58e5cd250633059

**Note** The `<inheritAccess>` tag comes before the `<autoAssign>` tag in each range (this is out-of-order with respect to the sequence.

### Details
Currently when a server returns values in a `<sequence>` that are out-of-order zeep throw an exception: https://gist.github.com/nmaludy/d21a6cc266c844298a0465f5d1da47e4

This is fine because technically this is not valid XML according to the WSDL schema. However, when i pass in `strict=False` to the client i get all of the out-of-order elements dumped into a `_raw_elements` list on the object. https://gist.github.com/nmaludy/db792168e466e7626ea102569aead04d

### Thoughts
When passing `strict=False` would it be possible to handle out-of-order values in the sequence and assign them to the correct keys in the dict instead of dumping them to `_raw_elements`?

mvantellingen:
If this is an AXIS soap server then this seems be a duplicate of #493
Can you confirm that? (It seems only the axis servers are doing this completely wrong)

nmaludy:
@mvantellingen I don't believe so, this is a Men&Mice IPAM server running on Windows IIS.

mvantellingen:
Okay. I agree that this should be done better. I'll look into it

Jitsusama:
FYI; I opened up issue #596 before recognizing that it's a duplicate of this and #493. This is biting me really hard with my AXIS server. Feel free to take a look at my ticket for another example of how sorted children tags are causing breakage.

nmaludy:
@mvantellingen If you can point me in the right direction i really wouldn't mind fixing it. I tried going through the code, but it it's my first time looking at it and i got a little lost in all of the `lxml` details.

nmaludy:
@mvantellingen I'm still interested in helping out here, if you can help point me in the right direction!

dankrause:
[...]

PR #880, opened 2018-12-04
Add xsd_ignore_sequence_order flag

### LINKED ISSUE #493: Implement an IS_AXIS compability flag to work with broken AXIS soap servers
The AXIS soap server is broken and still seems to be used a lot. So we can now pass strict=false to zeep to workaround it somewhat by not parsing unexpected elements, an IS_AXIS flag might just process xsd:sequence as xsd:all when it is set

dankrause:
Fixes #493
Based on https://github.com/macrojames/python-zeep/commit/e2dc96ef53b6548a79e13e6ff935fb5e06d1739d by @macrojames

### DIFF
--- a/src/zeep/settings.py
+++ b/src/zeep/settings.py
@@ -34,6 +34,10 @@ class Settings(object):
      approach to add http headers for specific calls.
     :type extra_headers: list
 
+    :param xsd_ignore_sequence_order: boolean to indicate whether to enforce sequence
+     order when parsing complex types. This is a workaround for servers that
+     don't respect sequence order.
+    :type xsd_ignore_sequence_order: boolean
     """
     strict = attr.ib(default=True)
     raw_response = attr.ib(default=False)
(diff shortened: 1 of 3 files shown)


dankrause:
This should also close #499

mvantellingen:
This is for an axis soap server i presume? I have the idea to add a dialect setting to zeep, a bit like the csv module where you can specify that the dialect is axis and that this is enabled by in that case. 

We can probably even auto detect it based on the http headers when dialect is set to auto

dankrause:
While a "dialect" setting makes a lot of sense, and sounds like a great feature, I suggest implementing it as a sort of collection of changes to the default settings, and also allow those settings to be toggled individually.

In this case, I would not have known that this is a problem common with axis servers, as l had never heard of that software until I found the open issues on this module.

I recommend that you add the individual settings such as the one in this pull request, so that they can be discovered easily, and then build the dialect feature to change groups of them at a time in a convenient way.

mvantellingen:
Okay fair enough, I think this will indeed solve a number of issues for people! Thanks!

PR #883, opened 2018-12-06
Added a test for multiple message parts.

apollo13:
This pull request consists only of a test which will currently fail. I have added a WSDL file which has two `wsdl:part`s as part of a `wsdl:message`. 

It seems to fail due to
https://github.com/mvantellingen/python-zeep/blob/713c779db7699b3c13fc9b1948eb3c6cfe32d828/src/zeep/wsdl/messages/soap.py#L422-L439
which results in double wrapping of the soap message.

I am not sure what to fix there though, if I remove the wrapping it fails somewhere else :)

### DIFF
--- a/tests/test_wsdl_messages_document.py
+++ b/tests/test_wsdl_messages_document.py
@@ -332,6 +332,78 @@ def test_serialize():
     assert_nodes_equal(expected, serialized.content)
 
 
+def test_serialize_multiple_parts():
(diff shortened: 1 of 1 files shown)


mvantellingen:
Should be fixed now. Not too happy with the current implementation, but good enough for now i suppose

apollo13:
Indeed, thank you!

PR #888, opened 2018-12-14
Fixed Chameleon schema support and added support for an edgecase with unqualified elements.

apollo13:
This partially reverts the recently introduced support for chameleon
schemas in the sense that it doesn't consider the targetNamespace of the
WSDL as default namespace for the schema. Sadly it is valid (while ugly
and not widely used) to have schemas with no namespace.

Secondly this commits allows refering to those unqualified elements if
an explicit empty <import/> (sic) is found. This seems to be mostly in
line with https://www.w3.org/TR/2012/REC-xmlschema11-1-20120405/#src-resolve
(if I read it correctly).


mvantellingen:
Thanks for the follow up. Reading the link you provided and since it seems to work for you i'm going to merge this :-)

Thanks again for the work

PR #890, opened 2018-12-17
Try modern namespace for WSA first

zanchey:
The WSDL shipped by a service I use has a different namespace for the action attribute, which means that the attributes don't get parsed correctly. I'm still getting my head around all the WS Addressing standards, but as far as I can tell the current zeep code uses a namespace from a superseded recommendation (https://www.w3.org/TR/2006/CR-ws-addr-wsdl-20060529/) while the current recommendation (https://www.w3.org/TR/2007/REC-ws-addr-metadata-20070904/) suggests something else entirely.

This commit tries the current standard first before falling back to the superseded one.

Example schema (Apache licensed)
[pcehr_schema.zip](https://github.com/mvantellingen/python-zeep/files/2686336/pcehr_schema.zip)

Script to demonstrate usage:
```python3
#!/usr/bin/env python3

from zeep import Client
from datetime import datetime
from lxml import etree

client = Client("pcehr_schema/wsdl/External/B2B_GetAuditView.wsdl")
serv = client.create_service("{http://ns.electronichealth.net.au/pcehr/svc/GetAuditView/1.1}getAuditViewServiceSOAP12Binding", 'uri:someaddress')
e = client.create_message(serv, 'getAuditView', datetime.now(), datetime.now())

print(etree.tostring(e, pretty_print=True, xml_declaration=True, encoding='utf-8').decode())
```

### DIFF
--- a/src/zeep/wsdl/parse.py
+++ b/src/zeep/wsdl/parse.py
@@ -12,6 +12,7 @@ from zeep.wsdl import definitions
 NSMAP = {
     'wsdl': 'http://schemas.xmlsoap.org/wsdl/',
     'wsaw': 'http://www.w3.org/2006/05/addressing/wsdl',
+    'wsam': 'http://www.w3.org/2007/05/addressing/metadata',
 }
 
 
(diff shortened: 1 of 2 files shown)


mvantellingen:
neat, thanks!

PR #899, opened 2019-01-14
Log unexpected elements when strict is False

ThreePinkApples:
Setting strict to False can be a risky move, but possibly a necessary workaround. It is then good to have all the unexpected elements to be logged, to help with debugging any issues that would have been caught with strict=True


ThreePinkApples:
Was browsing my GitHub history and noticed this. Rebased with master :)

ThreePinkApples:
Stumbled upon this old PR, rebased it now, but it has no relevance for me anymore (not worked with Python for 5 years now...)

PR #905, opened 2019-01-18
Added availability to make async plugins

smilingDima:
This patch makes availability to use asyncio library inside plugins.
Functions "_create" and "process_reply" are simply copied from master branch with adding async/await keywords.
Inside plugins.py added verifing if the function is async or not.


smilingDima:
Added test for async_plugin

PR #919, opened 2019-02-22
Annotate regular expression as an r'string'

lucaswiman:
@mvantellingen This PR annotates a string containing `\[` as an `r'string'`. This prevents an invalid escape sequence warning on newer versions of Python 3:
```
DeprecationWarning: invalid escape sequence \[
    match = re.match('([^\[]+)', array_type)
```


mvantellingen:
merged, thanks

PR #926, opened 2019-03-05
[FIX] avoid error when the WSDL has a non binding port

umiphos:
This error can be replicated with this gist
https://gist.github.com/umiphos/543e92eef8c26aad44262bd55ba73baa
And this URL
https://e-factura.sunat.gob.pe/ol-ti-itcpfegem/billService?wsdl

You just make a new Client(URL) and you get the error from before.

Now, as you can see, that WSDL has a definition without bindings, that
makes an index error when zeep tries to get the services and bindings.
To fix that we apply this patch, in case of not bindings, we should just
warn about it and continue with the rest of the document.

### DIFF
--- a/src/zeep/wsdl/definitions.py
+++ b/src/zeep/wsdl/definitions.py
@@ -121,7 +121,11 @@ class Binding(object):
         self._operations = {}
 
     def resolve(self, definitions):
-        self.port_type = definitions.get('port_types', self.port_name.text)
+        try:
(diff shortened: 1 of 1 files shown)


umiphos:
@erwaller if you can tell me how to fix the creation of the linux environment I can fix it, it tries to import a linux dependency to create the image but the reference does not exist(probably deprecated)

PR #931, opened 2019-03-09
Update test requirements

felixonmars:
Tests pass here with the updated versions.


mvantellingen:
> Tests pass here with the updated versions.

Thanks, just fixed it myself before seeing this pr

PR #942, opened 2019-03-24
fix(xop): respect RFC2392 when mapping href to cid

afeliciano:
[RFC link]( https://tools.ietf.org/html/rfc2392)

>  A "cid" URL is converted to the corresponding Content-ID message
   header [MIME] by removing the "cid:" prefix, converting the % encoded
   character to their equivalent US-ASCII characters, and enclosing the
   remaining parts with an angle bracket pair, "<" and ">"

### DIFF
--- a/src/zeep/wsdl/messages/xop.py
+++ b/src/zeep/wsdl/messages/xop.py
@@ -1,4 +1,5 @@
 import base64
+from six.moves.urllib.parse import unquote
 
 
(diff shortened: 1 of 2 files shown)


afeliciano:
This should also solve the problem reported in the following [PR](https://github.com/mvantellingen/python-zeep/pull/896)

mvantellingen:
Nice, thanks!

PR #950, opened 2019-04-12
WSSE: Don't create an empty timestamp node in the Header just so we can sign it!

3bh:
Resolves an issue I had connecting to a service that doesn't require the Timestamp element.


mvantellingen:
Thanks!

PR #957, opened 2019-04-30
Allow customizing the digest and signature method during wsse signing. Fixes #616

### LINKED ISSUE #616: sha256 instead of sha1
Hi,

Is is possible to use sha256 for WSSE XML Request instead of sha1?
1. The version of zeep - 2.4

apollo13:


mvantellingen:
Awesome, thanks

PR #959, opened 2019-05-08
Sequence can be set as xsd.Nil

### LINKED ISSUE #503: Question: how to create empty element which is a choice with optional sequence
Hi, I'm not sure, it's a bug, or I'm just missing something.

zeep==2.2.0

This is a part of xsd:

Basically, there is choice between `Offer`/`ApartmentBuildingOwner`/`LivingHouseOwner`/`Organization`. For `LivingHouseOwner` there is a choice: `RegOrg` or `Ind` . And for `Ind` I have an option to make it empty element or complete with sequence.

It means, that I should be able to use (**option_1**):
```xml
<ns0:LivingHouseOwner>
    <ns0:Ind>
        <ns0:Surname>Pupkin</ns0:Surname>
    </ns0:Ind>
</ns0:LivingHouseOwner>
```
or this (**option_2**):
```xml
<ns0:LivingHouseOwner>
    <ns0:Ind/>
</ns0:LivingHouseOwner>
```

I have no problem with **option_1**:
```python
{'LivingHouseOwner': {'Ind': {'Surname': 'Pupkin'}}}
```

But cannot figure out, how to achieve **option_2**. I've tried this:
```python
{'LivingHouseOwner': {'Ind': {}}
```
but it ends with an error: `zeep.exceptions.ValidationError: Missing element Surname (importSupplyResourceContractRequest.Contract.Surname)`

These two are also no good:
```python
{'LivingHouseOwner': {'Ind': []}
{'LivingHouseOwner': {'Ind': None}
```
because they just end up with no `LivingHouseOwner` element in the message.

So, the question is, how can I achieve **option_2**?

smilingDima:
Added availability to set xsd.Nil to sequence that have min_occurs=0.
Fixes #503

### DIFF
--- a/src/zeep/xsd/types/complex.py
+++ b/src/zeep/xsd/types/complex.py
@@ -6,7 +6,7 @@ from itertools import chain
 from cached_property import threaded_cached_property
 
 from zeep.exceptions import UnexpectedElementError, XMLParseError
-from zeep.xsd.const import NotSet, SkipValue, xsi_ns
+from zeep.xsd.const import NotSet, SkipValue, Nil, xsi_ns
 from zeep.xsd.elements import (
     Any, AnyAttribute, AttributeGroup, Choice, Element, Group, Sequence)
 from zeep.xsd.elements.indicators import OrderIndicator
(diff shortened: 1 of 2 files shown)


mvantellingen:
Nice, thanks!

PR #960, opened 2019-05-09
support constraining facets

### LINKED ISSUE #513: Zeep dont handle restriction on simpleTypes?
1. The version of zeep: 2.2.0 (python 2.7)
2. The WSDL you are using
```
<xs:simpleType name="RollenEnum">
    <xs:restriction base="xs:string">
        <xs:enumeration value="VersichertePerson"/>
        <xs:enumeration value="Versicherungsnehmer"/>
        <xs:enumeration value="Beitragszahler"/>
        <xs:enumeration value="BezugsberechtigterImTodesfall"/>
        <xs:enumeration value="Ehepartner"/>
        <xs:enumeration value="GesetzlicherVertreter"/>
    </xs:restriction>
</xs:simpleType>
```
Hello there,

I'm only using the zeep library to parse WSDL files, so I don't use it as a real SOAP Client.

I need this to map it against own datamodels. However, I can't figure out if zeep handles restrictions in simpleTypes. The Type 'RollenEnum' is after parsing a 'String' BuiltinType. This is good in the first place. But I also need to map the restrictions like Enums and so on. Doesn't zeep support this or do I miss something?

I also tried to create Objects of this Type where I was able to input other Values than defined in the Enumeration. 

Another example would be this:
```
<xs:simpleType name="someType">
    <xs:restriction base="xs:integer">
        <minInclusive value="0"/>
        <maxInclusive value="100"/>
    </xs:restriction>
</xs:simpleType>
```
Again, if I create a Type of 'someType' like so: ``someType(120)`` I don't get any Exception or something although the Value isn't valid like the Schema said. 

So again my Question: Do I miss something or is this just not supported?

Thank you very much in advance for the help!

hirochachacha:
Added initial implementations. Some constraints are still missing.

Fixes #513

### DIFF
--- a/src/zeep/xsd/constraining_facets.py
+++ b/src/zeep/xsd/constraining_facets.py
@@ -1,0 +1,215 @@
+import decimal
+import keyword
+import re
+
+from zeep.exceptions import ValidationError
+from zeep.xsd.const import xsd_ns
+from zeep.xsd.types.builtins import (
+    AnyURI,
+    Base64Binary,
+    Boolean,
+    Date,
+    DateTime,
+    Decimal,
+    Double,
+    Duration,
+    Float,
+    HexBinary,
+    Notation,
+    QName,
+    String,
+    Time,
+    gDay,
+    gMonth,
+    gMonthDay,
+    gYear,
+    gYearMonth,
+)
+
+
+class tags(object):
+    pass
+
+
+for name in [
+    "restriction",
+    "length",
+    "minLength",
+    "maxLength",
+    "pattern",
+    "enumeration",
+    "whiteSpace",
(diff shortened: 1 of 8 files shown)


hirochachacha:
Done. I think most of features are now supported.

I haven't tackled some subtle points:
  * most of facets are not recommended to have multiple items.
    * For example, if there are two length constraints - length=1 and length=2 - , it's not clear to me how validation should work. Maybe this is unspecified behavior.
  * some facets support 'fixed' attribute which prevent further restriction.

Those problems are validation of XML code itself, rather than runtime validation.
Currently, this PR use permissive approach.

Issue #958, opened 2019-05-02
Import xsd from other folder fails


yokovaski:
I just found that with Zeep version 3.3.1 I cannot create a client from a wsdl that imports xsd from another path.

For example:
```
<xsd:import schemaLocation="xsd/common.xsd" />
```

Zeep tries to import this xsd from `https:/server.com/common.xsd`. When I use Zeep 3.3.0, the xsd is imported correctly.

yokovaski:
I just tested the fix that has been merged in to master and that fixes my problem. Thank you! 
When will there be a new release from master with this fix?

yokovaski:
Fix is in release 3.4.0

PR #962, opened 2019-05-10
Fix Incorrect absolute path resolution of XSD import

adambogocz:
Path was resolved relative to the root document, but if parsing was indirected by another include or import a wrong path was created. This issue was found by loading FIXM XSD files. 

This fix solves issue mvantellingen/python-zeep#958

### DIFF
--- a/src/zeep/xsd/visitor.py
+++ b/src/zeep/xsd/visitor.py
@@ -183,7 +183,7 @@ class SchemaVisitor(object):
         location = node.get("schemaLocation")
         if location:
             location = normalize_location(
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks, are you able to add a unittest for this so that it won't be broken again in the future?

adambogocz:
I wanted to do it right away, but when I saw test_xsd_visitor.py I was not sure about adding test case that needs filesystem access, since all tests in that unit were done on in-placed xml strings. Is it OK with you?

PR #966, opened 2019-05-21
 Solve the value of a long string, unicode or str type judgment

StevenXue:
unexpected string value, attr_value set to NotSet by default

### DIFF
--- a/src/zeep/xsd/types/complex.py
+++ b/src/zeep/xsd/types/complex.py
@@ -242,17 +242,17 @@ class ComplexType(AnyType):
         if value is None:
             value = {}
 
-        if isinstance(value, str) or isinstance(value, unicode):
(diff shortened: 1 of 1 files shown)


src/zeep/xsd/types/complex.py
```diff
@@ -247,7 +247,17 @@ def render(self, parent, value, xsd_type=None, render_path=None):
 
         # Render attributes
         for name, attribute in self.attributes:
-            attr_value = value[name] if name in value else NotSet
+            import sys
+            if sys.version_info < (3, 4):
+                if isinstance(value, str) or isinstance(value, unicode):
```
mvantellingen: an `isinstance(value, basestring)` should be enough here

StevenXue: ok, I retry to make new pull request

PR #998, opened 2019-08-28
Added abstract transport that allows to modify LOG of request/response

prietopa:
Test is skipped, because it's not the best test, and it is very hard to test log console output.


mvantellingen:
I don't believe this should be part of the core of zeep.  This can best be done via a package (either as plugin or custom transport)

PR #1001, opened 2019-09-06
Fix AttributeError: NoneType object has no attribute clone

brianjarita:


mvantellingen:
Thanks! Can you add a unittest?

PR #1006, opened 2019-09-13
Make tests compatible with pytest5

jonringer:
This is to allow your test suite to be ran with pytest5 :)


jonringer:
been 3 months with no activity, closing

PR #1009, opened 2019-09-17
Added an Elements factory to the Client

serafdev:
Added an element factory that does the same thing as the types factory. but for elements.

Basically this functionality just lets you access `Elements` more easily like this:
```python
elements.Item(title="The Item's Title")
```
Instead of the old:
```python
client.get_element("ns0:Item")(title="The Item's Title")
```

### DIFF
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -145,13 +145,26 @@ class Client(object):
 
         Example::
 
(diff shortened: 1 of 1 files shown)


serafdev:
Closing, IDK how SOAP works anymore

PR #1024, opened 2019-10-28
Drop support for python 3.3 & 3.4

apollo13:
Since both Python 3.4 & 3.3 are EOL and new lxml versions do not support 3.4 anymore, drop it from lxml.


apollo13:
@mvantellingen Hi, this PR would drop Python3.3 & 3.4 so the tests pass again. If you think it is very important to support 3.3 & 3.4 I'll try to come up with a fix :/

mvantellingen:
Seems good! I was thinking about a major version bump and going 3.5+ only (removing 2.x)


Thanks

PR #1027, opened 2019-11-05
added support for dates that are compliant with RFC3339

nofalx:
Many old API use this format for dates:
```
%Y-%m-%d %H:%M:%S
2019-11-04 00:00:00
```
This raise exception in zeep because the expected ISO format must include `T` instead of space:
```
%Y-%m-%dT%H:%M:%S
2019-11-04T00:00:00
```
this addition will automatically detect the case and add a `T` instead of space.

Note the first date format is compliant with `RFC3339` standards and is used by many old API providers
https://stackoverflow.com/a/34006233

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "/home/ahmad/github.com/10ten-me/python-zeep/env/bin/python3"
+}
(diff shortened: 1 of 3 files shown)


nofalx:
@mvantellingen

mvantellingen:
Thanks, This PR adds a vscode config file. Can you remove that?

mvantellingen:
merged myself

PR #1031, opened 2019-11-14
improved documentation of not-default binding with examples.

andreabisello:


slowdust:
I'm still struggling with this but maybe because I don't understand the concept behind it, my issue is that when I use WSDL url I have plenty of validations and elements to send (for instance in the header there are no mandatory elements even though I'm receiving unexpected keyword arguments message from signature header, this is giving me a bunch of unnecessary work because if I could use ?SOAP url just one element would be enough to achieve what I need.
Any practical example in how to use SOAP url instead of WSDL? Thanks in advance

mvantellingen:
Thanks

Issue #990, opened 2019-07-29
tornado 5 or later support


toddrme2178:
This package requires tornado 4 (`tornado>=4.0.2,<5`).  However, tornado 4 is unmaintained, and with Python 3.8 it will stop working altogether.  Further, other packages are dropping support for tornado 4 (such as most of the Jupyter stack) which makes it harder to install this package alongside them.  Tornado 5 (released March 2018) and Tornado 6 (released March 2019) both support python 3.8.  Would it be possible to add support one or both of these versions so that zeep will continue to run on Python 3.8?

PR #1034, opened 2019-12-10
test python 3.8

akeeman:
Shows that Zeep works with python 3.8.

Optional requirement tornado 4 does not support python 3.8 (https://github.com/mvantellingen/python-zeep/issues/990)


mvantellingen:
Thanks!

PR #1035, opened 2019-12-13
allow bytes-like object as password argument to UsernameToken

curtisgr:
This PR allows `UsernameToken` to take a bytes-like object as password argument which I believe is in line with the spec.

https://www.oasis-open.org/committees/download.php/13392/wss-v1.1-spec-pr-UsernameTokenProfile-01.htm

The spec says of `PasswordText`:

```The actual password for the username, the password hash, or derived password or S/KEY.```

I have a use-case that requires the password be SHA-1 before being concatenated, e.g.,

```# digest = Base64 ( SHA-1 ( nonce + created + SHA-1 ( password ) ) )```

The current implementation throws an `AttributeError` when calling `self.password.encode('utf-8')`. This PR checks to see whether `.encode` should be called on `self.password`.

### DIFF
--- a/src/zeep/wsse/username.py
+++ b/src/zeep/wsse/username.py
@@ -1,6 +1,7 @@
 import base64
 import hashlib
 import os
+import six
 
 from zeep import ns
 from zeep.wsse import utils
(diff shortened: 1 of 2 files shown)


mvantellingen:
Thanks!

PR #1058, opened 2020-02-11
Fix Wrong WSA Action with Reused InputMessages

TheJoCr:
Allow AbstractOperations which share the same InputMessage to reference different WSA Actions as well. Previously, whenever two AbstractOperations referenced the same InputMessage, whichever was declared second would overwrite the wsa_action on the former. Here, we properly segregate the all message objects to whatever abstract operation they are associated with.


TheJoCr:
Made a few changes - this should be in a good state now.

mvantellingen:
Thanks! Adding the test which previously failed and now works should be enough (trusting all the other available unittests for now)

PR #1065, opened 2020-03-04
Change NotImplemented to NotImplementedError

sezginacer:
As mentioned at [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError) & [NotImplemented](https://docs.python.org/3/library/constants.html#NotImplemented),`NotImplemented` is not exception class and can not be raised. `NotImplementedError` should be raised instead of `NotImplemented`. This pull request will fix this bug.


mvantellingen:
Thanks!

PR #1092, opened 2020-05-27
Avoid convert duration by isodate if argument is string type

gnseo:
When converting duration value to xmlvalue, if the duration value as argument is already a string type, then just return the string.
I'm working on a case that receives payload as JSON, and the duration values are string type, such as "PT1S".
If I pass those duration values to zeep, it produces "P%P".
So I made this change to xmlvalue function to avoid force converting.


mvantellingen:
Can you add a unittest?

PR #1096, opened 2020-06-08
Add ssl_context parameter in AsyncTransport

### LINKED ISSUE #1095: Cannot provide custom key/cert for client-side verification with AsyncTransport
1. zeep version: 3.2.0
2. WSDL: not relevant

There is currently no way to provide an SSL context for use by the underlying aiohttp.ClientSession. This prevents us from setting a custom key and certificate using AsyncTransport, which prevents us from using client-side verification.

We are currently working around this issue by subclass AsyncTransport but I feel it should be part of the main library. Required change would be minimal, I'll create a pull request.

lerela:
in order to provide a custom SSLContext for client-side verification in async mode (fixes #1095).

I'm unsure how to test this as this would require aioresponses to be able to distinguish SSLContext.

### DIFF
--- a/examples/async_client.py
+++ b/examples/async_client.py
@@ -1,4 +1,5 @@
 import asyncio
+import ssl
 import time
 
 import zeep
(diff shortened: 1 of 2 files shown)


mvantellingen:
aiohttp is dropped in zeep 4.0 (replaced by httpx)

PR #1097, opened 2020-06-11
RedisCache

BradleyKirton:
Hi @mvantellingen, first up thank you for the zeep library. I think this is an outstanding example of structuring a Python project and the code is excellent. I plan to use zeep in a project and would like to make use of Redis as a caching backend. So I thought I would add it and make a PR.

I have written some tests for the redis backend and made the redis dependency optional. I also added doc strings and provided a very brief mention of the backend in the docs. Please let me know if you are open adding this functionality and if there is something else that is required. I have tested this on Python 3.8.

### DIFF
--- a/src/zeep/plugins.py
+++ b/src/zeep/plugins.py
@@ -46,7 +46,7 @@ def apply_ingress(client, envelope, http_headers, operation):
     return envelope, http_headers
 
 
-class HistoryPlugin:
+class HistoryPlugin(Plugin):
     def __init__(self, maxlen=1):
         self._buffer = deque([], maxlen)
 


mvantellingen:
Hi @BradleyKirton,

Nice implementation. I think this should be published as a separate package (zeep-redis-cache?) to keep the base easier to maintain. Is that something you can do? If there are code changes required to make thjis possible then let me know

BradleyKirton:
Hi @mvantellingen, yes absolutely.

I created another base class in this PR `VersionedCacheBase` which both the sqlite and redis cache made use of. I can make a new PR with only that change and then create a separate package which makes use of that base class for the redis cache.

PR #1106, opened 2020-07-06
in aiohttp >= 3.3.0 Timeout is now ClientTimeout

jensbjorgensen:
Hi, with newer aiohttp the aiohttp.Timeout object is no more.  See:

https://docs.aiohttp.org/en/stable/client_quickstart.html#aiohttp-client-timeouts


mvantellingen:
aiohttp is replaced by httpx in zeep 4.0.0

PR #1114, opened 2020-07-20
Implemented zulu timestamp and hashed passwords before concatenation

shivshankardayal:
Some SOAP services want zulu timestamps with Z in timestamps and in password digests they may want password to be hashed before adding it to nonce and created.


fabianoengler:
+1 to this

mvantellingen:
Thanks! merged. Are you also able to add some docs for this?

mvantellingen:
It also misses unittests

PR #1116, opened 2020-07-21
Handle message parts with no element

### LINKED ISSUE #1115: wsdl:message with type rather than element is not correctly parsed
When` a message is defined with a part that is a type, for example:
```
<wsdl:message name="ProblematicResponse">
  <wsdl:part name="text" type="xsd:string" />
</wsdl:message>
```
Corresponding to a soap response of:
```
<s:Envelope xmlns:s="http://schemas.xmlsoap.org/soap/envelope/">
  <s:Header>
    ...
  </s:Header>
  <s:Body>
    <text>Some kind of interesting text</text>
  </s:Body>
</s:Envelope>
```
It fails to be correctly parsed. This is due to DocumentMessage._resolve_body() requiring an element to be present. I have a working fix for this which creates an element with the appropriate type, but I am unsure if this is actually the correct way to go. I am happy to put a test together and submit a PR.

da1910:
When no element is specified in a wsdl:part definition responses are not correctly handled. This PR handles this case and creates a dummy element with the appropriate type, allowing correct deserialization of responses.

Closes #1115

### DIFF
--- a/src/zeep/wsdl/messages/soap.py
+++ b/src/zeep/wsdl/messages/soap.py
@@ -454,8 +454,14 @@ class DocumentMessage(SoapMessage):
         else:
             sub_elements = []
             for part_name, part in parts.items():
-                element = part.element.clone()
-                element.attr_name = part_name or element.name
+                if part.element is not None:
+                    element = part.element.clone()
+                    element.attr_name = part_name or element.name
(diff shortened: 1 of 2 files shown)


mvantellingen:
Awesome :-)

PR #1117, opened 2020-07-24
WS-Security documentation: added UsernameToken with Timestamp token section & example

Alireza2n:
Hi,
I noticed [WS-Security documentation](https://docs.python-zeep.org/en/master/wsse.html) lacks a section on usage of `UsernameToken()` with `timestamp_token`.
I've added a section with an example to `wsse.rst`.

Thanks for all the hard work. :+1:


mvantellingen:
thanks!

PR #1131, opened 2020-08-17
Add VersionedCacheBase to cache module

BradleyKirton:
Hi @mvantellingen, I have moved the common methods for versioned caches into VersionedCacheBase and added some additional tests.

Let me know if there is anything else required for the PR to be merged.

Once this is merged I will go ahead and create a separate package `zeep-redis-cache`.


mvantellingen:
thanks!

PR #1132, opened 2020-08-17
In addition to Body, also sign Timestamp with the provided digest_method

ospeek:
While implementing a service with zeep, we found out that the optional Timestamp header was always signed using SHA1 instead of the supplied digest_method (SHA256 in our case). This fix will sign both the Body and the Timestamp using the provided digest_method.


mvantellingen:
Can you added a couple of unittests for this?

mvantellingen:
Merged #1201

PR #1139, opened 2020-09-01
Improve handling for windows file:// uris

### LINKED ISSUE #666: Wrong parsing of uri to local file on windows
1. Zeep version: 2.4.0
2. WSDL irrelevant

```python
import zeep
import pathlib
with tempfile.NamedTemporaryFile('wb') as f:
    zeep.Client(pathlib.Path(f.name).as_uri())
```
Gives an error

### LINKED ISSUE #1133: Transport load() function cannot deal with file:/// scheme
I have a use case where I need a local file as wsdl file

zeep's Transport load() function assumes that the file scheme will always only have two slashes, but that is not always the case. 

file:///C:/local/example/example.wsdl

will cause a failure.

da1910:
- Instead of reinventing the wheel, use a well-known requests plugin
- Treat uris according to RFC8189 appendix E.2
- Add new tests to verify this behaviour on windows and unix

Closes #1133 and closes #666, should allow windows users to properly specify and handle file uris, without breaking current correct support for unix systems.

Limitations
- Does not support non-local files like `file://anotherpc/c:/users/public/example.wsdl`
- Does not support the terse format with implicit "//" like `file:/usr/local/bin`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ install_requires = [
     "lxml>=3.1.0",
     "requests>=2.7.0",
     "requests-toolbelt>=0.7.1",
+    "requests-file>=1.5.1",
     "pytz",
 ]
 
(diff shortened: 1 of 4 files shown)


mvantellingen:
thanks :)

PR #1148, opened 2020-10-08
Remove useless import of aiohttp

cedk:
aiohttp is not a dependency so using zeep without it raise an
ImportError.


cedk:
Is it possible to get a bugfix release with this fix? This is because for now our CI is failing because it does not install aiohttp (not a dependency) but fails when importing zeep.

PR #1149, opened 2020-10-10
Fix AsyncTransport verify_ssl and proxy

adammillerio:
The conversion to using httpx for AsyncTransport brought with it a few bugs that make it unusable:

* Any proxy or verify_ssl setting provided was not passed to the wsdl_client, meaning that it was more or less impossible to get a useable client if verification needs to be disabled.
* Incorrect keyword arguments were used for verify_ssl (should be verify) and proxy (should be proxies).

This fixes these by passing any verify_ssl or proxy settings into the client constructors, so that they persist across all interactions. In addition, timeout settings were also moved into the client.

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -179,14 +179,18 @@ class AsyncTransport(Transport):
             raise RuntimeError("The AsyncTransport is based on the httpx module")
 
         self.cache = cache
-        self.wsdl_client = wsdl_client or httpx.Client()
-        self.client = client or httpx.AsyncClient()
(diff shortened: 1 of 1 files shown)


mvantellingen:
Good point! Thanks! Everything else working correctly?

adammillerio:
@mvantellingen Yup! So far so good with async zeep after fixing this up. I'll keep you posted if I hit anything strange along the way.

mvantellingen:
Good to hear, I've just released 4.0.0. Hoping that this (partial) async implementation is easier to maintain :)

Issue #1151, opened 2020-10-13
AsyncClient's context manager?


decaz:
zeep 4.0.0

`AsyncClient` has `__aexit__` method, but it doesn't have `__aenter__` (so its usage as context manager raises `AttributeError`).

Also I guess there should be "**a**close":
https://github.com/mvantellingen/python-zeep/blob/6c9019cdd5052899c08f31a68bb3c1d39bfdc027/src/zeep/client.py#L237-L238

PR #1163, opened 2020-11-04
Fix AsyncClient's context manager

psanlorenzo:
Fixes: https://github.com/mvantellingen/python-zeep/issues/1151


ovnicraft:
@psanlorenzo rebase and unify your commits and left just one.

PR #1169, opened 2020-11-23
Added support for namespaces in SOAP 11 Fault node

vascohenriques:
Hello everyone,

# Change Description
Change to fix issue with namespace assertion in SOAP 11 Binding of Faults.
`fault_node.find(name, namespaces=fault_node.nsmap)`

PR #1171, opened 2020-12-02
fault node is not found when namespaced

mfyuce:


apollo13:
I think this can be closed in favor of #1169 which also has tests.

mvantellingen:
Just merged #1169 so this can be closed

PR #1193, opened 2021-03-01
Fix settings context manager leaks on exception

qvik-olli:
Fix settings context manager does not do proper cleanup when an exception is raised.


mvantellingen:
Thanks!

Issue #1175, opened 2020-12-16
Authentication with AsyncTransport


papamauro:
I'm trying to perform HTTP authentication with AsyncTransport using the following code:

`session = Session()`
`session.auth = HTTPBasicAuth(user, password)`
`transport = AsyncTransport(session=session, cache=None)`
`client = AsyncClient('http://my-endpoint.com/production.svc?wsdl', transport=transport)`

But it returns me the following error:

`AttributeError: 'AsyncTransport' object has no attribute 'session'`

How can I correctly perform authentication for async calls? Thank you!

jensenja:
Hi,

Synchronous transport is handed by the `requests` module, but asynchronous transport is handled by the `httpx` module.

To perform HTTP basic auth with `httpx`, you need to create an `httpx.AsyncClient` object and pass the basic auth parameters in to that object - then you need to pass that object in as the `client` parameter to `AsyncTransport`, like this:

```
#!/usr/bin/env python

import httpx
import zeep
from zeep.transports import AsyncTransport

USER = 'username'
PASSWORD = 'password'

aclient = zeep.AsyncClient(
    "http://localhost:8080/webservices/Eval?wsdl",
    transport=AsyncTransport(client=HAAC)
)
```

Hope this helps.

papamauro:
Hello, thank you very much!
I tried this code but authentication does not seem to work properly. I always get "Unauthorized" while the same credentials work fine for the synchronous zeep-only code as provided in the [documentation
](https://docs.python-zeep.org/en/master/transport.html#http-authentication)
I also tried to pass authentication credentials manually using the _headers_ option in the `httpx `AsyncClient, but nothing changed.

jensenja:
Hi,

I'm not sure where to point you next, as my example code works for me. Have you tested against a different SOAP server to isolate the issue from that side?

tonal:
Full worked code:
```python
async def aio_main():
  client = httpx.AsyncClient(auth=(login, password))
  wsdl_client = httpx.Client(auth=(login, password))
  client = zeep.AsyncClient(
    wsdl=wsdl_url, transport=AsyncTransport(client=client, wsdl_client=wsdl_client))
  client.wsdl.dump()

asyncio.run(aio_main())
```

Skeen:
I have stumbled upon the same issue.

[...]

PR #1196, opened 2021-03-08
docs: Add information about http auth for async client

cdvv7788:
This solution covers a pretty common use case. It was solved in https://github.com/mvantellingen/python-zeep/issues/1175 but it should be documented somewhere. I tried it and it worked.


mvantellingen:
Thanks!

PR #1201, opened 2021-03-22
Pass digest method when signing timestamp node

colm29:
This allows a digest method other than the SHA1 default to be used when signing the timestamp node.

I see now that there is an existing similar PR (#1132 ) but I'll add a couple of tests to this one if that one is not moving.


colm29:
Modified existing timestamp tests to validate digest method has been passed

mvantellingen:
thanks!

Issue #1168, opened 2020-11-18
create_service for AsyncClient


jenson2486:
Version: zeep 4.0.0
WSDL: 'http://www.dataaccess.com/webservicesserver/numberconversion.wso?WSDL'

Hi guys,
i am starting developing a soap interface based on zeep. Our application works mostly async, so I try to use the AsyncClient. Additional I test the interface against a mock, placed in soap_ui. For using the mock, it is necessary to define the endpoint to the mock, so I used the create_service method.

```
client = AsyncClient(wsdl='http://www.dataaccess.com/webservicesserver/numberconversion.wso?WSDL')
service = client.create_service('{http://www.dataaccess.com/webservicesserver/}NumberConversionSoapBinding',
                                'http://[replaced]:8088/mockNumberConversionSoapBinding')
loop = asyncio.new_event_loop()
print(loop.run_until_complete(service.NumberToDollars(2)))
```
The expception log:
```
Traceback (most recent call last):
  File ".\tests\minimal_examples\soap.py", line 46, in <module>
    print(loop.run_until_complete(service.NumberToDollars(2)))
  File "C:\Users\wernerj\ee-interface-webservice\.tox\dev\lib\site-packages\zeep\proxy.py", line 46, in __call__
    return self._proxy._binding.send(
  File "C:\Users\wernerj\ee-interface-webservice\.tox\dev\lib\site-packages\zeep\wsdl\bindings\soap.py", line 135, in send
    return self.process_reply(client, operation_obj, response)
  File "C:\Users\wernerj\ee-interface-webservice\.tox\dev\lib\site-packages\zeep\wsdl\bindings\soap.py", line 177, in process_reply
    if response.status_code in (201, 202) and not response.content:
AttributeError: 'coroutine' object has no attribute 'status_code'
sys:1: RuntimeWarning: coroutine 'AsyncTransport.post_xml' was never awaited
```

After some research, the reason is the create_service method, defined in the sync client, which returns a sync ServiceProxy object, but exected will be a AsyncServiceProxy object.

Is my assumption right?

BR Jens

william-davis-od:
Your assumption is correct, I just had the same issue. I resolved in my project by creating a custom class that extends the AsyncClient class and overrides the create_service method to return an AsyncServiceProxy: 

```
   def create_service(self, binding_name, address):
        """Create a new AsyncServiceProxy for the given binding name and address.

        :param binding_name: The QName of the binding
        :param address: The address of the endpoint

[...]

PR #1202, opened 2021-03-23
Add create_service to AsyncClient

jan-23:
Fixes: #1168
create_service needs to be overridden to be supported for the AsyncClient.


lxop (commented):
A more elegant solution might be to add a class variable `ServiceProxyClass`, which on the `Client` class is set to `ServiceProxy` and on the `AsyncClient` class is set to `AsyncServiceProxy`. Then both of these methods (`bind` and `create_service`) can be updated to have `return self.ServiceProxyClass(...)` and the `AsyncClient` class can be reduced to simply

```python
class AsyncClient(Client):
    ServiceProxyClass = AsyncServiceProxy

````

PR #1204, opened 2021-03-24
Add more typing details

Dreamsorcerer:


src/zeep/transports.py
```diff
@@ -4,7 +4,7 @@
 from urllib.parse import urlparse
 
 import requests
-from requests import Response
+from requests import Response as Response
```
Dreamsorcerer, line 7: This allows using the Response type without directly depending on requests (which avoids any issues if the underlying library were ever changed).

Used like:
```
from zeep.transports import Response
resp: Response = client_with_raw_response.service.Foo()
```

Issue #1205, opened 2021-03-26
Error adding ONVIF camera (XMLParseError: No namespace defined for 'ter')


slydiman:
It seems this module is used in https://github.com/hunterjm/python-onvif-zeep-async and finally in Home Assistant Core https://github.com/home-assistant/core

I have reported this issue https://github.com/home-assistant/core/issues/45513 few months ago without any progress.

Probably you can help to investigate and fix the problem.

The component ONVIF inside Home Assistant Core contains the following requirements:
    "onvif-zeep-async==1.0.0",
    "WSDiscovery==2.0.0",
    "zeep[async]==4.0.0"

I got the error 
`zeep.exceptions.XMLParseError: No namespace defined for 'ter' ('ter:InvalidArgVal')`
trying to add the camera Wanscam K21 or Wanscam K54 in Home Assistant.
The full traceback is
```
2021-01-25 03:57:41 DEBUG (MainThread) [homeassistant.components.onvif] Setting up the ONVIF device management service
2021-01-25 03:57:41 DEBUG (MainThread) [homeassistant.components.onvif] Retrieving current device date/time
2021-01-25 03:57:41 DEBUG (MainThread) [homeassistant.components.onvif] Device date/time: 2021-01-24 23:57:42+00:00 | System date/time: 2021-01-24 23:57:41.482773+00:00
2021-01-25 03:57:44 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry Cam1 - xx:xx:xx:xx:xx:xx for onvif
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 236, in async_setup
    result = await component.async_setup_entry(hass, self)  # type: ignore
  File "/usr/src/homeassistant/homeassistant/components/onvif/__init__.py", line 77, in async_setup_entry
    if not await device.async_setup():
  File "/usr/src/homeassistant/homeassistant/components/onvif/device.py", line 105, in async_setup
    self.capabilities = await self.async_get_capabilities()
  File "/usr/src/homeassistant/homeassistant/components/onvif/device.py", line 252, in async_get_capabilities
    pullpoint = await self.events.async_start()
  File "/usr/src/homeassistant/homeassistant/components/onvif/event.py", line 83, in async_start
    await self.async_renew()
  File "/usr/src/homeassistant/homeassistant/components/onvif/event.py", line 154, in async_renew
    await self._subscription.Renew(termination_time)
[...]

PR #1206, opened 2021-04-01
Fix WSA action bug which breaks Home Assistant ONVIF component

slydiman:
Fix 400 Bad Request response to RenewRequest because Action and other addressing data is missing in the header.
https://github.com/mvantellingen/python-zeep/issues/1205

### DIFF
--- a/src/zeep/wsdl/bindings/soap.py
+++ b/src/zeep/wsdl/bindings/soap.py
@@ -81,7 +81,7 @@ class SoapBinding(Binding):
             if not options:
                 options = client.service._binding_options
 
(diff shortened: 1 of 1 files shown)


slydiman:
Ping.

sag3ll0:
Yes its break HA. This fix is needed

mvantellingen:
Merged, although i do wonder about if this is the correct way. We are now always adding WSA afaik

mvantellingen:
Just reverted it again, needs more thinking (if you always want WSA you could just list it in the plugins)

PR #1213, opened 2021-04-16
The function Client.wsdl.dump() was modified to allow to put the print output to a file or other stream.

### LINKED ISSUE #1212: Modify Client.wsdl.dump() to allow to put the print output to a file or other stream
Currently the Client.wsdl.dump() just has the hability to print on screen.
It would be nice if we can have the choice to direct the print message to a file or the an io.StringIO() like as the bult-in function print already permits.
This implementation it is not complex and gives more flexibility to the function dump().

renatodamas:
The function Client.wsdl.dump() was modified to allow to put the print output to a file or other stream.

To the function interface was add a default parameter _**output=sys.stdout**_ that keeps the current usability and permits flexility to change the output.

close #1212

### DIFF
--- a/src/zeep/wsdl/wsdl.py
+++ b/src/zeep/wsdl/wsdl.py
@@ -7,6 +7,7 @@ from __future__ import print_function
 
 import logging
 import operator
+import sys
 import os
 import typing
 import warnings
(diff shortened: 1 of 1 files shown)


src/zeep/wsdl/wsdl.py
```diff
@@ -106,43 +107,43 @@ def load(self, location):
     def __repr__(self):
         return "<WSDL(location=%r)>" % self.location
 
-    def dump(self):
-        print("")
-        print("Prefixes:")
+    def dump(self, output=sys.stdout):
+        print("", file=output)
```
frost-nzcr4, line 111: You can use local redefinition of print to avoid to set multiple `file=output`:

```python
from functools import partial

print = partial(print, file=output)

print("")
print("Prefixes:")
```

and `file` (as used in print) is better than `output`.

renatodamas: @frost-nzcr4 I will performe the changes and submit another pull request.

PR #1240, opened 2021-06-30
fix data type date parsing

frNNcs:


leranp:
it fix my problem , thanks

mvantellingen:
I think only the `import` change is needed right?

frNNcs:
I remove the line
``` python
if len(value) == 10:
    value += "T00:00:00"
elif (len(value) == 19 or len(value) == 26) and value[10] == " ":
    value = "T".join(value.split(" "))
```
too, that conflicted with the date parsing

mvantellingen:
This PR is way to vague, no clear idea what it solves

PR #1244, opened 2021-07-16
Use platformdirs instead of appdirs

felixonmars:
Use the better maintained `platformdirs` instead of `appdirs`.


mvantellingen:
👍

PR #1250, opened 2021-08-15
Get rid of six

sbraz:
Hi,
It looks like 1ce4e7f2f12461064783ff4dc82f2641172663aa missed one instance where `six` is used.


mvantellingen:
thanks!

Issue #1255, opened 2021-09-03
Bug parsing multiple repeated choice elements in a sequence.


steven-deprez:
When trying to parse a sequence that contains more than one choice, each with a maxOccurs > 1, zeep fails to parse valid xml.
The error message tells you that the first choice does not appear in the sequence.

I made a pull request fixing the issue together with a test case.
The test case reproduces the problem.

PR #1256, opened 2021-09-03
fix name of choice elements in sequence (issue #1255)

steven-deprez:
Fix + added test case for issue #1255 

Currently zeep fails to parse a schema like the following, telling you that he expects either an "a" or a "b".
            <?xml version="1.0"?>
            <schema xmlns="http://www.w3.org/2001/XMLSchema"
                    xmlns:tns="http://tests.python-zeep.org/"
                    targetNamespace="http://tests.python-zeep.org/"
                    elementFormDefault="qualified">
                <element name="container">
                    <complexType>
                        <sequence>
                            <choice maxOccurs="2">
                                <element name="a" type="string"/>
                                <element name="b" type="string"/>
                            </choice>
                            <choice maxOccurs="2">
                                <element name="c" type="string"/>
                                <element name="d" type="string"/>
                            </choice>
                        </sequence>
                    </complexType>
                </element>
                <element name="a" type="string" />
                <element name="b" type="string" />
            </schema>

The problem is that the value of the second choice overwrites the value of the first one.

### DIFF
--- a/src/zeep/xsd/elements/indicators.py
+++ b/src/zeep/xsd/elements/indicators.py
@@ -615,7 +615,7 @@ class Sequence(OrderIndicator):
             for elm_name, element in self.elements:
                 try:
                     item_subresult = element.parse_xmlelements(
-                        xmlelements, schema, name, context=context
+                        xmlelements, schema, elm_name, context=context
                     )
                 except UnexpectedElementError:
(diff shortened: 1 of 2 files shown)


steven-deprez:
fixes #1255

PR #1265, opened 2021-10-09
Close internal session when transport is deleted

cedk:
When using the default session, a "ResourceWarning: unclosed" is raised because the requests session is not closed explicitly.


mvantellingen:
Thanks :-)

PR #1273, opened 2021-11-11
Use stdlib functools.cached_property if available

mgorny:
Python 3.8+ provides a functools.cached_property in the stdlib that is
thread-safe, i.e. equivalent to threaded_cached_property.  Use it
instead of adding third-party dependencies whenever available.


mvantellingen:
Thanks 👍

mgorny:
You're welcome!

PR #1283, opened 2021-12-23
Fix empty SOAPAction header

davidolrik:
SOAPAction header should never be "None".

When `self.operation.soapaction` is `None` it was stringifyed into `"None"` when it should be `""`.

This fixed the following error when calling a Soap 1.1 service on Camunda:

```xml
<soap:Envelope xmlns:soap="http://schemas.xmlsoap.org/soap/envelope/">
  <soap:Body>
    <soap:Fault>
      <faultcode>soap:Server</faultcode>
      <faultstring>The given SOAPAction None does not match an operation.</faultstring>
    </soap:Fault>
  </soap:Body>
</soap:Envelope>
```


jdury:
+1 same issue, and fix works like a charm

PR #1300, opened 2022-03-17
Add BinaryMemorySignature

kjagiello:
`MemorySignature` is really useful when you already have the key and certificate in the memory, but I found myself wanting to do the same for `BinarySignature` and realized that there was no equivalent for it. This PR introduces `BinaryMemorySignature`.


mvantellingen:
Awesome, tanks

PR #1302, opened 2022-03-18
Mark imported symbols as public for Pyright compatibility

ernstwi:
This is a fix for [the following Pyright rule](https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface):

> Imported symbols are considered private by default. If they use the “import A as A” (a redundant module alias), “from X import A as A” (a redundant symbol alias), or “from . import A” forms, symbol “A” is not private unless the name begins with an underscore.

The docs state that "typecheckers like Pyright" use this rule, but I don't know what standard this is based on TBH.

I ran `make format` after adding the aliases, which split all compound imports into separate lines.


mvantellingen:
Seems duplicate with #1326 Closing this for now (although this was older)

Issue #1319, opened 2022-07-25
Add AsyncTransport to the __init__ file


parfeniukink:
You have added **AsyncClient** to thee `__init__` file for doing `from zeep import AsyncClient` but it is impossible with **AsyncTransport** as far as it it not in the `src/zeep/__init__.py`

mvantellingen:
This was done since not all Python versions supported the async keyword. Can be done now

PR #1320, opened 2022-07-25
Add AsyncTransport to the zeep module import scope

parfeniukink:
https://github.com/mvantellingen/python-zeep/issues/1319


mvantellingen:
added

PR #1321, opened 2022-07-28
feat: add response verification

mmdanny89:
Add new feature: Response verification. with certificate.


mvantellingen:
I can merge this if test cases are added

mmdanny89:
test case added, check it. thanks.

PR #1330, opened 2022-08-08
Allow to pass in a pre parsed Document

michaelheyvaert:
Although the zeep client has a caching mechanism,
every client instantiation still incurs the overhead of processing the cached WSDL document. 

This PR allows a user to pass in a pre-parsed Document so it can be used by multiple Clients.

I wasn't sure if I should add an example to the docs, since this is a more advanced use case,
but I would be glad to add this if it is useful.

### DIFF
--- a/src/zeep/client.py
+++ b/src/zeep/client.py
@@ -38,7 +38,7 @@ class Factory:
 class Client:
     """The zeep Client.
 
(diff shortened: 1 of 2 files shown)


mvantellingen:
Thanks!

PR #1339, opened 2022-09-15
Allow CDATA text 

sysnux:
We needed to send embedded XML as CDATA in the request.
Without this modification, the embedded XML was escaped, now it is transmitted unmodified.


mvantellingen:
Thanks!

PR #1342, opened 2022-09-30
Use explicit reexport

Dreamsorcerer:
If a user disable implicit reexport in Mypy, then they will get errors when trying to use anything in the zeep module.
https://mypy.readthedocs.io/en/stable/config_file.html#confval-implicit_reexport

This PR disables implicit reexport and updates the code to not introduce errors for users.

Also, just fixed a bunch of type errors while I was there.

Currently (mypy 0.981) need to run `mypy --enable-recursive-aliases`, but in a future version that will be enabled by default and can just run `mypy`.

### DIFF
--- a/examples/trafficvance_apikey.py
+++ b/examples/trafficvance_apikey.py
@@ -3,7 +3,7 @@ from zeep import Client, xsd
 API_KEY_TEST = 'YOUR_OWN_API_KEY'
 WSDL_TEST = 'https://apitest.trafficvance.com/?v3=system.wsdl'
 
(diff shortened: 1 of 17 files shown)


mvantellingen:
Hi, thanks for the PR. I actually prefer the `__all__` approach as that feels more how Python always does it. Is it possible to do that instead?

Dreamsorcerer:
Yes, see the examples:
https://mypy.readthedocs.io/en/stable/command_line.html?highlight=reexport#cmdoption-mypy-no-implicit-reexport

PR #1350, opened 2022-11-20
Fix to deserializer if result type does not have a length

holstebroe:
An exception will be thrown if a bool is returned from a SOAP service call. `deserialize` soap.py will ask for the length of the result body, but it may not be allowed to take len on some result body types. Added check if length is valid and returns the body directly if it is not. I haven't tested with any other types, such as integers.

### DIFF
--- a/src/zeep/wsdl/messages/soap.py
+++ b/src/zeep/wsdl/messages/soap.py
@@ -116,6 +116,8 @@ class SoapMessage(ConcreteMessage):
             return result
 
         result = result.body
+        if not hasattr(result, '__len__'): # Return body directly if len is allowed (could indicated valid primitive type).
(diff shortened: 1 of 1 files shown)


mvantellingen:
Thanks!

src/zeep/wsdl/messages/soap.py
```diff
@@ -116,6 +116,8 @@ def deserialize(self, envelope):
             return result
 
         result = result.body
+        if not hasattr(result, '__len__'): # Return body directly if len is allowed (could indicated valid primitive type).
```
frost-nzcr4, line 119: It seems the comment should be _Return body directly if len is **NOT** allowed_.

PR #1351, opened 2022-12-09
WSA plugin: allow custom prefix for MessageID.

### LINKED ISSUE #975: Optionally remove "urn" prefix from wsa:MessageID
Hi,

I'm trying to connect an application to the Dutch Chamber of Commerce (KvK) dataservice using python 3.7.3 and `zeep==3.4.0`. This appears to be a bit of a pain. 

The wsdl for this service is available via http://schemas.kvk.nl/contracts/kvk/dataservice/catalogus/2015/02/. No problems there yet. 

wsa is added automatically, probably due to data found in the wsdl. The `wsa` headers are added twice when I add the `WsAddressingPlugin` via the `plugins` parameter at initialization of the `Client` instance. I'll try to add a reproducer later on.

There's a requirement in the docs saying the service is unable to process messages that contain a `MessageID` which is prefixed by `urn`. I think a lot of developers would be happier when the service would be able to process messages containing that prefix instead of documenting this issue, but that's out of my control. Here's an excerpt in Dutch, but I'm guessing that's not an issue for you ;)
> LET OP: sommige IDE tooling zoals bijvoorbeeld Microsoft Visual Studio voegt “urn:” voor “uuid:”. De Handelsregister Dataservice controles zullen een request met deze syntax afwijzen!
(https://www.kvk.nl/sites/aansluitendataservice/index.html#/webservice-werkingsprincipes)

Is there a way to remove the `urn` prefix from the `MessageID`? Do I need to override the plugin somewhere?

Merinorus:
Fix #975

### DIFF
--- a/src/zeep/wsa.py
+++ b/src/zeep/wsa.py
@@ -13,8 +13,9 @@ WSA = ElementMaker(namespace=ns.WSA, nsmap={"wsa": ns.WSA})
 class WsAddressingPlugin(Plugin):
     nsmap = {"wsa": ns.WSA}
 
-    def __init__(self, address_url: str = None):
+    def __init__(self, address_url: str = None, message_id_prefix: str = "urn:uuid:"):
         self.address_url = address_url
+        self.message_id_prefix = message_id_prefix
 
     def egress(self, envelope, http_headers, operation, binding_options):
         """Apply the ws-addressing headers to the given envelope."""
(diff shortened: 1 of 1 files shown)


bartnv:
Any chance of getting this merged? It would solve a problem for me as well (I'm also trying to use the KvK Dataservice with python-zeep, like other participants in issue https://github.com/mvantellingen/python-zeep/issues/975). Currently I need to keep a modified version of wsa.py around to make this work.

PR #1365, opened 2023-03-16
Update index.rst with a new example

alephmelo:
The current service seems to be offline for quite a while.


mvantellingen:
I'm not sure if we can use that domain instead. It would be best to host our own example wsdl service. Need to think about this

PR #1369, opened 2023-03-27
Retry request once on server disconnect

bdraco:
fixes https://github.com/home-assistant/core/issues/90199

[datatracker.ietf.org/doc/html/rfc2616#section-8.1.4](https://datatracker.ietf.org/doc/html/rfc2616#section-8.1.4)

   A client, server, or proxy MAY close the transport connection at any
   time. For example, a client might have started to send a new request
   at the same time that the server has decided to close the "idle"
   connection. From the server's point of view, the connection is being
   closed while it was idle, but from the client's point of view, a
   request is in progress.

   This means that clients, servers, and proxies MUST be able to recover
   from asynchronous close events. Client software SHOULD reopen the
   transport connection and retransmit the aborted sequence of requests
   without user interaction so long as the request sequence is
   idempotent (see section 9.1.2). Non-idempotent methods or sequences
   MUST NOT be automatically retried, although user agents MAY offer a
   human operator the choice of retrying the request(s). Confirmation by
   user-agent software with semantic understanding of the application
   MAY substitute for user confirmation. The automatic retry SHOULD NOT
   be repeated if the second sequence of requests fails.

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -1,6 +1,7 @@
 import logging
 import os
 from contextlib import contextmanager
+from typing import Any, Callable, TypeVar
 from urllib.parse import urlparse
 
 import requests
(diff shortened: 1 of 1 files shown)


bdraco:
I solved this with a decorator downstream

PR #1370, opened 2023-04-15
Handle datetimes that use a - instead of T

bdraco:
Makes "2016-03-04-21:14:42.123456" parsable


bdraco:
I solved this by subclassing the DateTime type downstream

PR #1384, opened 2023-07-05
[FIX] multiple nested complex types expecting "any" and receiving etree._Element input

felipezago:
There is an issue when the WSDL consists of multiple nested complex types and requires an "any" input, and you attempt to provide the input as an `etree._Element` object.

On the `render` method it was always expected that the value received was a dictionary. However, in this specific scenario, the "any" element is represented as an etree._Element, resulting in the element_value receiving the `NotSet` value. As a consequence, this situation triggers an exception: zeep.exceptions.ValidationError: Missing element for Any.

The WSDL where this issue was encountered is structured as follows:

image

### DIFF
--- a/src/zeep/xsd/elements/indicators.py
+++ b/src/zeep/xsd/elements/indicators.py
@@ -236,18 +236,20 @@ class OrderIndicator(Indicator, list):
 
         self.validate(values, render_path)
 
+        child_path = render_path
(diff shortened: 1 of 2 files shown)


mileo (approved):
LGTM

ygcarvalh (approved):

PR #1421, opened 2024-07-02
drop python<=3.7 support

kloczek:
According to https://endoflife.date/python python 3.7 has been EOSed 27 Jun 2023.
Filter all code over `pyupgracde --py38-plus`.


mvantellingen:
Thanks!, merged it without the string formatting changes as I don't believe that makes the codebase more readable

kloczek:
Thank you 👍

PR #1437, opened 2024-10-14
Use f-strings in most places

number492:
This is a simple refactor to replace most string formatting methods with f-strings. Found via pylint.

F-strings are considered more readable and more performant than other string formatting options. This PR does not attempt to fix any mistakes, it should produce functionally identical code.


mvantellingen:
I do think f-strings are generally a good idea, but in this case i don't feel the code becomes more readable. So i have to close this

Issue #1420, opened 2024-05-28
RFE: use zoneinfo instead of pytz


kloczek:
Switch to standard `zoneinfo` module.
Below may help  https://github.com/pydantic/pydantic-core/commit/fd262933

deronnax:
I can do that. @mvantellingen would you accept a PR doing this change ?

mvantellingen:
That would be great!

deronnax:
I did the work in #1442

PR #1442, opened 2024-10-28
remove pytz

deronnax:
To keep it simple, I made it so it imply to drop support for python 3.8, [which is already OEL](https://devguide.python.org/versions/).
Though, I still have this error I am not sure about:

```
tests/test_xsd_builtins.py:158 (TestDateTime.test_xmlvalue)
'2016-03-04T21:14:42+00:00' != '2016-03-04T21:14:42Z'

Expected :'2016-03-04T21:14:42Z'
Actual   :'2016-03-04T21:14:42+00:00'

<Click to see difference>


self = <tests.test_xsd_builtins.TestDateTime object at 0x1036aa090>

    def test_xmlvalue(self):
        instance = builtins.DateTime()
        value = datetime.datetime(2016, 3, 4, 21, 14, 42)
        assert instance.xmlvalue(value) == "2016-03-04T21:14:42"
    
        value = datetime.datetime(2016, 3, 4, 21, 14, 42, tzinfo=timezone.utc)
>       assert instance.xmlvalue(value) == "2016-03-04T21:14:42Z"
E       AssertionError: assert '2016-03-04T21:14:42+00:00' == '2016-03-04T21:14:42Z'
E         
E         - 2016-03-04T21:14:42Z
E         ?                    ^
E         + 2016-03-04T21:14:42+00:00
E         ?                    ^^^^^^

tests/test_xsd_builtins.py:165: AssertionError
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ license = { text = "MIT" }
 authors = [
     { name = "Michael van Tellingen", email = "michaelvantellingen@gmail.com" }
 ]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 classifiers = [
     "Development Status :: 5 - Production/Stable",
     "License :: OSI Approved :: MIT License",
@@ -27,7 +27,6 @@ dependencies = [
     "requests>=2.7.0",
     "requests-toolbelt>=0.7.1",
     "requests-file>=1.5.1",
-    "pytz",
 ]
 
 [project.urls]
--- a/src/zeep/cache.py
+++ b/src/zeep/cache.py
@@ -8,7 +8,6 @@ from contextlib import contextmanager
 from typing import Dict, Tuple, Union
 
 import platformdirs
-import pytz
 
 # The sqlite3 is not available on Google App Engine so we handle the
 # ImportError here and set the sqlite3 var to None.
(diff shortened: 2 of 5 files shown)


deronnax:
OK I know why, it's because `isodate` has a formatting logic that do not recognize python's UTC tzinfo objects: their `.dst()` method return `None`, isodate expects a `timedelta(0)` to consider them UTC.
* either we use isodate's tzinfo everywhere instead of python's one, but that would be a move backward
* either we need to find something so we can get the UTC datetime formatting to have a trailing `Z` for UTC datetime, because it's not doable out of the box with python standard datetimes.

JuneStepp:
> OK I know why, it's because `isodate` has a formatting logic that do not recognize python's UTC tzinfo objects: their `.dst()` method return `None`, isodate expects a `timedelta(0)` to consider them UTC.
> 
>     * either we use isodate's tzinfo everywhere instead of python's one, but that would be a move backward
> 
>     * either we need to find something so we can get the UTC datetime formatting to have a trailing `Z` for UTC datetime, because it's not doable out of the box with python standard datetimes.

See this commit from my fork: https://github.com/mvantellingen/python-zeep/commit/c16230e3a4d9a4b38949ac364e44ef56b61239a7

JuneStepp:
> Wow, that's a good work, maybe better than mine. Mind giving your work for #1420 ?

#1443 should do it. zoneinfo isn't actually needed, so I didn't change the Python version requirements.

deronnax:
closing because #1443 is superior.

Issue #1446, opened 2024-11-29
Incompatibilty with httpx


decaz:
Since [version 0.28.0 of httpx](https://github.com/encode/httpx/releases/tag/0.28.0) `The deprecated proxies argument has now been removed.`:

https://github.com/mvantellingen/python-zeep/blob/41a5115774b1807735dd1ddc6e08a36ca82ce094/src/zeep/transports.py#L186-L195

TB99930:
If anyone is experiencing this problem, simply fix your version of `httpx`, using an exact version pre-0.28 or the qualifier `<0.28` until the attached PR is merged.

For example: `httpx>=0.26,<0.28` or `httpx==0.27.2`

kostyaten:
When will the update be?

PrieJos:
Hi @ba1dr 

Could you please at least tell us whether there is any plan to fix this incompatibility issue or not?

Thanks
Jose M. Prieto

ba1dr:
> Hi [@ba1dr](https://github.com/ba1dr)
> 
> Could you please at least tell us whether there is any plan to fix this incompatibility issue or not?
> 
> Thanks Jose M. Prieto

I literally have no idea, as I am just the same user of this library as you. Looking at the past releases timeline I can only conclude that the author is busy and this library is semi-abandoned.

Perhaps eventually this will be fixed, but at the moment the best we can do is either override this issue with one of the examples above or make a fork and fix it for yourself.

PrieJos:
Hi @ba1dr 

Oops, sorry for that. I don’t know why I came to the conclusion you were the developer. My apologies.

> Perhaps eventually this will be fixed, but at the moment the best we can do is either override this issue with one of the examples above or make a fork and fix it for yourself.

Fully agree with you. Let’s see.

Cheers
José M Prieto

masaccio:
> If anyone is experiencing this problem, simply fix your version of `httpx`, using an exact version pre-0.28 or the qualifier `<0.28` until the attached PR is merged.
> 
> For example: `httpx>=0.26,<0.28` or `httpx==0.27.2`

That doesn’t work for Home Assistant for example which requires latest httpx. The code from @ba1dr works well and it’s what I am doing in my HA integration.

PR #1447, opened 2024-12-04
Update proxy argument in httpx Client/AsyncClient

aschollmeier-gcmlp:
Fixes https://github.com/mvantellingen/python-zeep/issues/1446

Ref: https://github.com/encode/httpx/blob/master/CHANGELOG.md#0260-20th-december-2023

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -183,15 +183,17 @@ class AsyncTransport(Transport):
 
         self._close_session = False
         self.cache = cache
+        proxy_kwarg_name = "proxy" if httpx.__version__ >= "0.26.0" else "proxies"
+        proxy_kwargs = {proxy_kwarg_name: proxy}
         self.wsdl_client = wsdl_client or httpx.Client(
             verify=verify_ssl,
-            proxies=proxy,
             timeout=timeout,
+            **proxy_kwargs,
         )
         self.client = client or httpx.AsyncClient(
             verify=verify_ssl,
-            proxies=proxy,
             timeout=operation_timeout,
+            **proxy_kwargs,
         )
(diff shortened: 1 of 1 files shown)


TB99930:
This looks good to me. Just needs @mvantellingen to merge when he gets a chance. Releasing this soon is important because anyone installing `zeep[async]` currently has a completely broken experience without this change.

src/zeep/transports.py
```diff
@@ -183,15 +183,17 @@ def __init__(
 
         self._close_session = False
         self.cache = cache
+        proxy_kwarg_name = "proxy" if httpx.__version__ >= "0.26.0" else "proxies"
```
shadchin, line 186: wrong check version, example
```
>>> "0.3.0" >= "0.26.0"
True
```
need to do something like this
```
>>> import packaging.version
>>> packaging.version.parse("0.3.0") >= packaging.version.parse("0.26.0")
False
```

aschollmeier-gcmlp: Good catch, thank you. I updated the version check.

imperiuse:
> This looks good to me. Just needs @mvantellingen to merge when he gets a chance. Releasing this soon is important because anyone installing `zeep[async]` currently has a completely broken experience without this change.

At least one person is waiting for this—me. =)
 I encountered this problem today... =(

wipeautcrafter:
We are waiting for this too: it would be nice if this was included in the next release!

aschollmeier-gcmlp:
@mvantellingen, just nudging on this. It looks like the linked issue is causing some pain points in the user base as people upgrade their httpx dependencies.

Sreesanth46:
@mvantellingen 🙏

PR #1454, opened 2025-02-18
fix(transports): allow headers extension with AsyncTransport

streambinder:
Currently, when passing a custom (wsdl) client with arbitrary headers set, they get overwritten only to set the User-Agent.
Let's make sure the User-Agent only gets added/modified rather.


streambinder:
Hey @mvantellingen, got any chance to look at this one?

PR #1462, opened 2025-05-21
Fixed wrong proxy keyword (httpx)

smorrin28:
Reverted PR #1149 since HTTPX now uses the keyword "proxy" instead of "proxies" (https://www.python-httpx.org/advanced/proxies/) to create any proxy values, this always throws an error when creating an AsyncTransport object if using the latest httpx version.

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -185,12 +185,12 @@ class AsyncTransport(Transport):
         self.cache = cache
         self.wsdl_client = wsdl_client or httpx.Client(
             verify=verify_ssl,
-            proxies=proxy,
+            proxy=proxy,
             timeout=timeout,
         )
         self.client = client or httpx.AsyncClient(
             verify=verify_ssl,
-            proxies=proxy,
+            proxy=proxy,
             timeout=operation_timeout,
         )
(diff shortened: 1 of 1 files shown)


Silverdoses:
Thanks for the PR. I ran into this issue today as well:

```python
TypeError: Client.__init__() got an unexpected keyword argument 'proxies'
```

For additional context: the `proxies` keyword argument was deprecated in earlier versions of `httpx` and was [removed](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0280-28th-november-2024) entirely in version `0.28.0`

This fix will solve issues when upgrading to latest version of `httpx` as dependency for other packages, like in my case.

smorrin28:
As a temporary fix I just directly created the wsdl client and passed it as an argument:

```python3
httpx_client = ...
wsdl_client = httpx.Client(verify=True, proxy=None, timeout=300)
client = AsyncClient(self._base_url, transport=AsyncTransport(client=httpx_client, wsdl_client=wsdl_client))
```

justin-ow:
This PR can be dropped. It was addressed in https://github.com/mvantellingen/python-zeep/pull/1447

PR #1476, opened 2025-11-26
fix: preserve attribute values when extending complexType

### LINKED ISSUE #991: Zeep setting defined valuesas 'NotSet'
Please keep the issues that real bug reports only. Please request help/support on a relevant channel like stackoverflow. This issue will most likely be closed.

swozny13:
When a complexType with only attributes was extended by another complexType, the attributes were being rendered as "NotSet" instead of their actual values. This occurred because a placeholder element was unnecessarily created during type extension, causing the base type to be rendered twice - once with correct values and once with empty values that overwrote the correct ones.

Fixes #991

### DIFF
--- a/src/zeep/xsd/types/complex.py
+++ b/src/zeep/xsd/types/complex.py
@@ -447,6 +447,8 @@ class ComplexType(AnyType):
 
         elif self._element or base_element:
             element = self._element or base_element
+        elif isinstance(base, ComplexType):
+            element = None
         else:
             element = Element("_value_1", base)
 
--- a/tests/test_xsd_complex_types.py
+++ b/tests/test_xsd_complex_types.py
@@ -422,3 +422,107 @@ def test_ignore_sequence_order():
 
     response = elm.parse(node[0], schema)
     assert response.Baz.id == 3
+
+
+def test_extension_with_attributes_preserves_values():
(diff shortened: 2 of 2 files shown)


swozny13:
**Zeep version: 4.3.1**

**Example schema**

```xml
<?xml version="1.0"?>
<schema xmlns="http://www.w3.org/2001/XMLSchema"
        targetNamespace="http://example.com/test"
        xmlns:tns="http://example.com/test"
        elementFormDefault="unqualified"
        attributeFormDefault="unqualified">

  <complexType name="BaseType">
    <attribute name="attr1" type="integer" use="required"/>
    <attribute name="attr2" type="string" use="required"/>
  </complexType>

  <complexType name="ExtendedType">
    <complexContent>
      <extension base="tns:BaseType"/>
    </complexContent>
  </complexType>

  <element name="item" type="tns:ExtendedType"/>
</schema>
```

**Reproduction**

```python
item = client.get_type('tns:ExtendedType')
obj = item(attr1=10, attr2="test")
print(etree.tostring(render(obj)))
```

**Before fix**

```xml
<item attr1="NotSet" attr2="NotSet"/>
```

**After fix**

```xml
<item attr1="10" attr2="test"/>
```

PR #1479, opened 2025-12-27
POST query string parameters - transport level

mouyang:
This is required to attach query strings to URLs which is used by some APIs to route SOAP messages on the backend.

Sample API that requires this - https://developer.ebay.com/api-docs/user-guides/static/make-a-call/using-soap.html#parameters


mouyang:
Issue #1496
Discussion #1477

PR #1480, opened 2025-12-27
pass operation from binding to transport

mouyang:
This passes the operation name to the Transport.  The API I'm working with needs this at the transport level in HTTP header for the raw XML version, or the query string in the SOAP version.

Sample XML API that requires this (X-EBAY-API-CALL-NAME) - https://developer.ebay.com/api-docs/user-guides/static/make-a-call/using-xml.html#headers

Sample SOAP API that requires this (callname parameter) - https://developer.ebay.com/api-docs/user-guides/static/make-a-call/using-soap.html#parameters

### DIFF
--- a/src/zeep/transports.py
+++ b/src/zeep/transports.py
@@ -54,7 +54,7 @@ class Transport:
             get_version()
         )
 
-    def get(self, address, params, headers):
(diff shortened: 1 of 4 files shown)


mouyang:
Issue #1497 
Discussion #1477

PR #1483, opened 2026-02-25
Fix/various fixes

swozny13:
When a complexType with only attributes was extended by another complexType, the attributes were being rendered as NotSet instead of their actual values. This occurred because a placeholder element was unnecessarily created during type extension, causing the base type to be rendered twice - once with correct values and once with empty values that overwrote the correct ones.

Add Signature class that reads key and certificate from files and allows verifying SOAP response signatures with a separate certificate different from the one used for signing.

### DIFF
--- a/src/zeep/wsse/signature.py
+++ b/src/zeep/wsse/signature.py
@@ -53,6 +53,8 @@ class MemorySignature:
         password=None,
         signature_method=None,
         digest_method=None,
+        verify_response_signature=True,
+        response_cert_data=None,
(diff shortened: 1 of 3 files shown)


mohosy (commented):
Quick review pass on 'Fix/various fixes': changedFiles=3, additions=119, deletions=1.\n\nPlease confirm regression coverage for the touched paths (especially edge/error handling), and consider adding a brief note for any behavior-visible changes.

Issue #1419, opened 2024-05-28
WSSE signature plugin for SecurityTokenReference


jaspercram:
I need to connect to a SOAP server that utilizes a WSSE authentication method different from the one available in Zeep. Unfortunately, I lack the knowledge and time to create a pull request for this plugin . However, I will share the code here in case it is useful to someone else. I apologize for not being able to provide more detailed information on how to use this plugin. Please feel free to ignore it or close this issue if necessary.

```python
class SignaturePlugin(zeep.Plugin):
    def egress(self, envelope, http_headers, operation, binding_options):
        # Insert filename here
        key = xmlsec.Key.from_file(XXXXXXX, xmlsec.constants.KeyDataFormatPem)
        soap_env = zeep.utils.detect_soap_env(envelope)

        # Create the Signature node.
        signature = xmlsec.template.create(envelope, xmlsec.Transform.EXCL_C14N, xmlsec.Transform.RSA_SHA256)

        # Insert the Signature node in the wsse:Security header.
        security = get_security_header(envelope)
        security.insert(0, signature)

        # Perform the actual signing.
        ctx = xmlsec.SignatureContext()
        ctx.key = key
        _sign_node(ctx, signature, envelope.find(QName(soap_env, 'Body')))
        timestamp = security.find(QName(zeep.ns.WSU, 'Timestamp'))
        if timestamp is not None:
            _sign_node(ctx, signature, timestamp)
        ctx.sign(signature)

        # Place the X509 data inside a WSSE SecurityTokenReference within
        # KeyInfo. The recipient expects this structure, but we can't rearrange
        # like this until after signing, because otherwise xmlsec won't populate
        # the X509 data (because it doesn't understand WSSE).
        key_info = xmlsec.template.ensure_key_info(signature)
        sec_token_ref = etree.SubElement(key_info, QName(zeep.ns.WSSE, 'SecurityTokenReference'))
        # Create the KeyIdentifier node.
        key_identifier = etree.SubElement(
            sec_token_ref,
            etree.QName(zeep.ns.WSSE, 'KeyIdentifier'),
            EncodingType='http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-soap-message-security-1.0#Base64Binary',
            ValueType='http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-x509-token-profile-1.0#X509SubjectKeyIdentifier',
        )
        # Insert Key Identifier here
[...]

Issue #1363, opened 2023-03-02
Example for Generating Binary Security Token Key Identifier Signatures


slrwin:
Dear Developer,

Could you please add some examples on how to create a Signature with Binary Security Token as Key Identifier.

Thanks in advance!!

Issue #1357, opened 2023-01-13
4.2.1: pytest errors `cannot load crypto library for xmlsec.`


kloczek:
I'm packaging your module as an rpm package so I'm using the typical PEP517 based build, install and test cycle used on building packages from non-root account.
- `python3 -sBm build -w --no-isolation`
- because I'm calling `build` with `--no-isolation` I'm using during all processes only locally installed modules
- install .whl file in </install/prefix>
- run pytest with $PYTHONPATH pointing to sitearch and sitelib inside </install/prefix>
- build is performed in env which is cut off from access to the public network (pytest is executed with `-m "not network"`)

I'm no sure what could be wron in my build env. If may I ask for some hints 🤔 
pytest shows as well some `PytestUnknownMarkWarning` warnings 

Here is pytest output:
<details>

```console
+ PYTHONPATH=/home/tkloczko/rpmbuild/BUILDROOT/python-zeep-4.2.1-2.fc35.x86_64/usr/lib64/python3.8/site-packages:/home/tkloczko/rpmbuild/BUILDROOT/python-zeep-4.2.1-2.fc35.x86_64/usr/lib/python3.8/site-packages
+ /usr/bin/pytest -ra -m 'not network'
==================================================================================== test session starts ====================================================================================
platform linux -- Python 3.8.16, pytest-7.2.0, pluggy-1.0.0
rootdir: /home/tkloczko/rpmbuild/BUILD/python-zeep-4.2.1, configfile: pyproject.toml, testpaths: tests
plugins: anyio-3.6.2, httpx-0.21.0, requests-mock-1.10.0
collected 466 items / 3 errors

========================================================================================== ERRORS ===========================================================================================
_______________________________________________________________________ ERROR collecting tests/test_wsse_signature.py _______________________________________________________________________
tests/test_wsse_signature.py:9: in <module>
    from zeep import ns, wsse
../../BUILDROOT/python-zeep-4.2.1-2.fc35.x86_64/usr/lib/python3.8/site-packages/zeep/wsse/__init__.py:2: in <module>
    from .signature import BinarySignature, MemorySignature, Signature
../../BUILDROOT/python-zeep-4.2.1-2.fc35.x86_64/usr/lib/python3.8/site-packages/zeep/wsse/signature.py:20: in <module>
    import xmlsec
E   xmlsec.Error: (1, 'cannot load crypto library for xmlsec.')
[...]

Issue #1318, opened 2022-07-14
xmlsec version mismatch error


israelcetinaa:
has someone the same error about the xmlsec version ?
![image](https://user-images.githubusercontent.com/76149563/179067877-3641c9c1-ecb4-44c4-8f92-65d7f3072389.png)

PR #1485, opened 2026-03-20
feat(wsse): pure-Python WS-Security signing (no xmlsec required)

martincollignon:
## Summary

Adds `zeep.wsse.crypto` — a pure-Python alternative to the existing xmlsec-based `zeep.wsse.signature` module. Uses the `cryptography` library instead of the C-based `xmlsec`, making installation straightforward on all platforms.

**Motivation**: We build [Landbruget.dk](https://github.com/landbruget-dk), a Danish agricultural data transparency project. Integrating with Denmark's VetStat SOAP API for antibiotic usage data required WS-Security features that zeep's current xmlsec-based module doesn't support — and installing `xmlsec` across CI/CD and developer machines was a constant pain point. We ended up writing ~300 lines of manual XML signing. This PR extracts that into a clean, general-purpose module that benefits everyone.

## What's new

| Feature | Current (xmlsec) | New (crypto) |
|---------|------------------|--------------|
| Sign Body + Timestamp | ✅ | ✅ |
| Sign UsernameToken | ❌ | ✅ |
| Sign BinarySecurityToken | ❌ | ✅ |
| Sign arbitrary elements | ❌ | ✅ (`extra_references`) |
| Inclusive namespace prefixes | ❌ | ✅ (per-reference) |
| PKCS#12 key loading | ❌ | ✅ |
| PEM key loading | ✅ | ✅ |
| Configurable digest algorithm | Partial | ✅ (SHA1/SHA256/SHA384/SHA512) |
| Configurable signature algorithm | Partial | ✅ (RSA-SHA1/SHA256/SHA384/SHA512) |
| No C library dependency | ❌ | ✅ |
| Signature verification | ✅ | ✅ |
| KeyIdentifier styles | ❌ | ✅ (ThumbprintSHA1, SKI) |
| Security header layout control | ❌ | ✅ (xmlsec_compatible, etc.) |
| Timestamp freshness validation | ❌ | ✅ |
| Certificate validity validation | ❌ | ✅ |

## Usage

```python
from zeep.wsse.crypto import CryptoSignature, CryptoBinarySignature

# Drop-in replacement for wsse.Signature
sig = CryptoSignature("key.pem", "cert.pem")

# BinarySecurityToken variant
sig = CryptoBinarySignature("key.pem", "cert.pem")

# PKCS#12 support
sig = CryptoBinarySignature.from_pkcs12("cert.p12", b"password")

# Sign extra elements + custom C14N prefixes (e.g. for government SOAP APIs)
sig = CryptoBinarySignature(
    "key.pem", "cert.pem",
    sign_username_token=True,
    sign_binary_security_token=True,
    digest_method="http://www.w3.org/2001/04/xmlenc#sha256",
    inclusive_ns_prefixes={"Body": ["wsse", "ds"]},
)

# Works with Compose
from zeep.wsse import Compose
from zeep.wsse.username import UsernameToken
wsse = Compose([UsernameToken("user", "pass"), sig])
```

Install: `pip install zeep[crypto]`

## Related issues

- #1357 (xmlsec installation failures)
- #1419 (SecurityTokenReference plugin)
- #1428 (signed parts)
- #1363 (BinarySecurityToken examples)
- #1318 (xmlsec version mismatch)

## Test plan

- [x] 38 tests covering all features (signing, verification, algorithm combos, PKCS12, inclusive prefixes, key identifiers, header layout, timestamp validation, Compose integration, edge cases)
- [x] No changes to existing `signature.py` — fully additive
- [x] Graceful fallback if `cryptography` not installed (classes set to `None` in `__init__.py`)
- [x] `crypto` optional dependency added to `setup.py` extras

🤖 Generated with [Claude Code](https://claude.com/claude-code)


taybin:
I've been using some of this code myself.  It generally works, but I did have to make some modifications.  Most importantly, the `load_der_x509_certificate` import is unused and I need to add an implementation for verifying signatures from the server's BinarySecurityToken.  I also had to explicitly add Timestamps to the Header.  So a couple of oversights, but overall this has been very helpful.

martincollignon:
Thanks for the feedback. I pushed a follow-up commit (`6c2ae06`) that addresses the oversights you called out:

- `load_der_x509_certificate` is no longer unused; it is now used to parse the DER X.509 certificate embedded in a response `wsse:BinarySecurityToken`.
- Response verification from `BinarySecurityToken` is supported via explicit opt-in: `verify(envelope, use_binary_security_token=True)`. The default `verify(envelope)` path still uses the configured certificate, so existing behavior is preserved.
- Added `timestamp_token` support to the crypto signature classes, matching the existing `UsernameToken(timestamp_token=...)` pattern. `sign_timestamp=True` still means “sign the timestamp if present,” and the new option lets the crypto plugin append one before signing.
- Added tests for BST-based verification success, wrong-cert default failure, tampered body digest failure, missing/malformed BST failures, appended timestamp signing, and no-timestamp behavior.

Focused verification passed locally:

```bash
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .venv/bin/python -m pytest tests/test_wsse_crypto.py
# 43 passed
```

taybin:
These changes seem to work, but I'm not sure how to set `use_binary_security_token=True` on CryptoMemorySignature.verify().  Isn't that `verify()` called by zeep itself?  I think this needs to be set at construction time on the object, and not passed in as an argument.

I also don't think passing in a timestamp at context construction time will work, as the timestamp will soon be out of date, and I believe we keep using the same security context.

taybin:
Thank you for your work on this.  The BST change looks good.  I think there is still an issue with the timestamp just being set once at wsse context creation time.  Shouldn't the timestamp be created inside CryptoMemorySignature.apply()?

something like:

```python
    def _make_timestamp_token(self, envelope: etree._Element) -> None:
        from datetime import timedelta

        security = get_security_header(envelope)  # type: ignore[no-untyped-call]
        now = datetime.now(UTC)
        timestamp = etree.SubElement(security, _wsu("Timestamp"))
        ensure_id(timestamp)  # type: ignore[no-untyped-call]
        created = etree.SubElement(timestamp, _wsu("Created"))
        created.text = now.strftime("%Y-%m-%dT%H:%M:%SZ")
        expires = etree.SubElement(timestamp, _wsu("Expires"))
        expires.text = (now + timedelta(seconds=300)).strftime("%Y-%m-%dT%H:%M:%SZ")

    def apply(
        self, envelope: etree._Element, headers: object
    ) -> tuple[etree._Element, object]:
        if (self.sign_timestamp):
            security = get_security_header(envelope)  # type: ignore[no-untyped-call]
            ts = security.find(_wsu("Timestamp"))
            if (ts is None):
                self._make_timestamp_token(envelope)

        self._sign(envelope)
        return envelope, headers
```

src/zeep/wsse/crypto.py
```diff
@@ -0,0 +1,1039 @@
...
+    etree.SubElement(ref, _ds("DigestMethod"), Algorithm=digest_uri)
+    dv = etree.SubElement(ref, _ds("DigestValue"))
+    dv.text = digest_value
+    return ref
+
+
+def _sign_envelope(
+    envelope: etree._Element,
+    private_key,
+    certificate,
```
taybin, line 264: This `certificate` argument is unused and can be removed.

PR #1443, opened 2024-11-04
Use standard library for date and time when possible

JuneStepp:
pytz dependency is removed. isodate is still needed for ISO durations and missing features in Python's ISO datetime parser.

closes #1442, closes #1420

PR #1498, opened 2026-07-13
Make sure datetime is a datetime

justmobilize:
In https://github.com/mvantellingen/python-zeep/pull/1443, all the special case code for pytz was removed since pytz isn't used anymore. Unfortunately it removed a case in builtins.DateTime where it would add in the time values if they were missing.

This causes issues if a user passes in a date to a datetime field as it get's sent up as an iso format date, not datetime and the receiving server may reject it.

This is a critical bug, and blocking updating to 4.3.3 with the security fix.

### DIFF
--- a/src/zeep/xsd/types/builtins.py
+++ b/src/zeep/xsd/types/builtins.py
@@ -149,6 +149,12 @@ class DateTime(BuiltinType):
         if isinstance(value, str):
             return value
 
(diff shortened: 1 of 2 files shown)


justmobilize:
@mvantellingen is this something you are willing to bring in and fix the breaking change?

PR #1499, opened 2026-07-23
fix: renderpath for type-extensions

m-hau:
The `child_path` variable can still be set from the attribute loop above, but then it has the wrong value. Without any attributes it is unset and causes an UnboundLocalError. Set it always to a proper value.

Fixes: 0c8a3fc4f9a8 ("Add support to pass base-types to type extensions")

### DIFF
--- a/src/zeep/xsd/types/complex.py
+++ b/src/zeep/xsd/types/complex.py
@@ -279,6 +279,7 @@ class ComplexType(AnyType):
             and not isinstance(value, (list, dict, CompoundValue))
         ):
             element = self.elements_nested[0][1]
+            child_path = render_path + [element.name]
             element.type.render(node, value, None, child_path)
             return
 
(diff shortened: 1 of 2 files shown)


victorwon2001 (commented):
Confirmed the fix against `1b7072c`: rendering a raw string through an attribute-free `simpleContent` extension raises `UnboundLocalError`, while `907f6c8` renders it correctly. The full suite passes locally (475 passed, 18 skipped), and Ruff is clean.

Could you add a regression test for that attribute-free path? A minimal schema with `<xsd:simpleContent><xsd:extension base="xsd:string"/></xsd:simpleContent>` and `element.render(..., "hello")` reproduces the failure on `main`. That would keep this one-line fix covered.

m-hau:
I added a test like requested, making sure it fails without the fix and passes with the fix.

While looking at the other tests to get an idea how to add my own, I noticed one of the tests in that file called `render_node(…)` twice without any apparent necessity or effect. It is probably a simple copy&paste error, since some tests call that function before the `expected` constant and others right before the `assert_nodes_equal`. I took the liberty of adding a second commit removing the second redundant call.

